Issue #250, opened 2026-07-01
[Medium] SignupLive handle_event("remove_authenticator") crashes on non-numeric index from a crafted websocket event


mkreyman:
**Severity:** Medium  |  **Confidence:** high  |  **Domain:** infra  |  **Class:** crash
**Finding ID:** `web-02` (loopctl pre-launch bug sweep)

### Affected code
- `lib/loopctl_web/live/signup_live.ex:187-193`
- `lib/loopctl_web/live/signup_live.ex:169-185`

### Trigger
A client on the public /signup LiveView pushes remove_authenticator with an arbitrary payload, e.g. %{"index" => "x"} (or a non-binary). The handler runs String.to_integer(index) directly on the attacker-controlled value.

### Observed behavior
String.to_integer("x") raises ArgumentError (and a non-binary index raises FunctionClauseError), crashing the LiveView process. Same class: attestation_error/2 pattern-matches %{"reason" => reason} and raises FunctionClauseError if a client sends the event without a "reason" key.

### Expected behavior
Malformed client events on a public LiveView should be ignored (no-op) rather than crashing the process; parse the index defensively (Integer.parse/1) and match a catch-all clause.

### Root cause
phx-value-index is normally server-set, but LiveView event payloads are fully client-controlled over the socket; there is no upstream validation of the event params before String.to_integer/1 / the strict pattern match. No RequireAuth guards this — the /signup route is deliberately public (router.ex:53-56).

### Reproduction
Connect to the /signup live socket and push_event("remove_authenticator", %{"index" => "abc"}); the LiveView crashes and reconnects.

### Suggested fix
Use Integer.parse/1 with a fallback (ignore invalid index and out-of-range), and add catch-all handle_event/3 clauses that no-op on unexpected payloads.

### Verification note
Independently verified (CONFIRMED). Verified: signup_live.ex:188-189 runs String.to_integer(index) directly on the client-controlled event payload — ArgumentError on non-numeric, FunctionClauseError on non-binary; attestation_error (line 170) hard-matches %{"reason" => reason}. /signup is deliberately public (router.ex:53-56 :public_signup live_session, :browser pipeline), so no auth guard prevents a crafted websocket event. Blast radius is the attacker's own LiveView process (auto-reconnects), so Medium is correct — unhandled-input crash on an unauthenticated surface, not a cross-user DoS.

---
_Filed from an automated pre-launch security & correctness sweep; each finding was independently re-verified against the cited code before filing._

[...]

Issue #234, opened 2026-07-01
[High] Signup rate limiter collapses to one global bucket (peer_data not in LiveView connect_info) — 5 page loads/hour lock out ALL tenant creation


mkreyman:
**Severity:** High  |  **Confidence:** high  |  **Domain:** infra  |  **Class:** exhaustion
**Finding ID:** `web-01` (loopctl pre-launch bug sweep)

### Affected code
- `lib/loopctl_web/live/signup_live.ex:38-74`
- `lib/loopctl_web/endpoint.ex:15-17`
- `lib/loopctl_web/router.ex:33-56`

### Trigger
Any user loads /signup. The rate limiter in mount/3 keys on the peer IP, but on the CONNECTED LiveView mount get_connect_info(socket, :peer_data) returns nil because the live socket's connect_info is only [session: ...] (no :peer_data), so peer_ip/1 falls to "unknown". check_rate("signup:unknown", 3_600_000, 5) is thus a single GLOBAL bucket, and it is incremented on every mount (page view), not only on completed signups.

### Observed behavior
After 5 connected /signup mounts per hour across the WHOLE platform, every subsequent visitor's websocket connect hits {:deny}, gets put_flash(:error, "Too many signup attempts") and push_navigate(to: "/"). Tenant creation — the only way to onboard — becomes unusable platform-wide at trivial traffic. The limiter is also not per-IP, so it cannot distinguish an abuser from legitimate users (all share the same bucket). (Disconnected mount uses Plug.Conn.get_peer_data, i.e. the raw transport peer = the Fly proxy IP behind the proxy, so it is likewise a shared global bucket, not the real client IP.)

### Expected behavior
Rate limit should be keyed on the real client IP (per-IP, e.g. 5 signups/hour/IP) and ideally count completed signups, not page views. peer_data must be registered in the live socket connect_info (and x_headers / RemoteIp-aware resolution used behind the Fly proxy) so distinct clients get distinct buckets.

### Root cause
[...]

PR #268, opened 2026-07-02
fix(signup): per-IP action-scoped rate limit + defensive event handling

mkreyman:
## Summary

Fixes two vulnerabilities in the public tenant-signup LiveView (`lib/loopctl_web/live/signup_live.ex`).

### web-01 (HIGH — resource exhaustion / self-DoS)

The signup rate limiter keyed on `get_connect_info(socket, :peer_data)`, which returns `nil` on the **connected** LiveView mount (the live socket's `connect_info` only carried `[session: ...]`). `peer_ip/1` fell back to `"unknown"`, so `check_rate("signup:unknown", ...)` collapsed into a **single global bucket** — and it was incremented on **every page view (mount)**, not on completed signups. After 5 connected `/signup` loads per hour across the whole platform, every subsequent visitor's websocket connect was denied, making tenant creation (the only onboarding path) unusable platform-wide at trivial traffic.

**Fix:**
- **Endpoint** (`endpoint.ex`): add `:peer_data` and `:x_headers` to the LiveView socket `connect_info` so the forwarded client IP is available on connect.
- Resolve the true client IP the same way the HTTP pipeline's `RemoteIp` plug does — via `RemoteIp.from/1` over the forwarded `x-forwarded-for` chain (populated by Fly's proxy), falling back to the direct websocket peer for non-proxied/local connections.
- Move the rate limit off `mount/3` (page views) and onto the abusable **`signup` submission** (tenant creation). Loading `/signup` never consumes the bucket.
- Bucket is **per real IP** (`signup:ip:<ip>`). For the genuinely-unknown-IP case (no forwarded header AND no peer), fall back to a **per-connection** key (`signup:session:<socket.id>`) — never a single shared bucket that would let one burst deny signup to everyone else.

### web-02 (MEDIUM — crash)

Malformed websocket events crashed the public SignupLive process:
- `remove_authenticator` called `String.to_integer/1` on the attacker-controlled `"index"` (`"x"` -> ArgumentError; non-binary -> FunctionClauseError).
- `attestation_error` pattern-matched `%{"reason" => reason}` and raised FunctionClauseError when sent without a `"reason"` key.

**Fix:**
- `remove_authenticator`: parse the index with `Integer.parse/1` and bounds-check against the current authenticator list; non-parseable / non-binary / out-of-range / missing indexes are no-ops.
- `attestation_error`: guard the happy path on a binary `"reason"`; add a no-op fallback for missing/non-binary reasons.
- Add a catch-all `handle_event/3` clause so any unknown/malformed frame (including `validate`/`signup` without their expected params) is ignored instead of crashing the process. Happy paths are unchanged.

## Tests

New coverage in `test/loopctl_web/live/signup_live_test.exs`:
- **web-01:** page views don't consume the bucket; the submission is checked once, keyed by the real forwarded IP (`signup:ip:203.0.113.7`); one IP hitting the limit does not deny a different IP; a genuinely-unknown IP falls back to a per-connection (`signup:session:...`) bucket, never a shared one.
- **web-02:** `remove_authenticator` with non-parseable / non-binary / out-of-range / missing index each leave the LiveView alive and are no-ops; a valid `remove_authenticator` still removes the right entry; `attestation_error` without a reason is a no-op; unknown events are ignored.

## Verification

`mix precommit` green: compile `--warnings-as-errors`, format check, `credo --strict`, deps.audit, dialyzer (0 real errors), full suite **3262 tests, 0 failures**.

Refs public issues #234 (web-01) and #250 (web-02).


mkreyman:
Follow-up commit `ab1d9fe` addresses all three gate findings (2 High, 1 Medium):

**FIX 1 (HIGH) — attestation path protected server-side.** `attestation_captured` now enforces, in-handler and before any crypto: the authenticator cap, a per-window verification budget (`signup:webauthn:<rate_key>`, 20/window), and an 8 KB pre-decode size cap — in that order, so capacity/size rejections don't draw down the budget. This can no longer be bypassed by skipping the `request_attestation` UI gate.

**FIX 2 (HIGH) — non-map crash class closed.** `request_attestation`, `attestation_captured`, and `remove_authenticator` now match a map-typed head with a `_params` no-op fallback (the old `with`/`else` caught non-matching results but `Map.get/3` still raised on a non-map first).

**FIX 3 (MEDIUM) — trusted-proxy IP resolution.** `plug RemoteIp` and the LiveView share one `:remote_ip_opts` config. SignupLive only trusts a forwarded client when the chain traversed a trusted proxy (a hop RemoteIp peels off the right); a direct public-peer connection keys on the unforgeable TCP peer and ignores the header; everything else (RFC 7239 `Forwarded` not in `:x_headers`, empty chain, shared proxy peer) fails safe to a **per-connection** bucket — never the shared peer.

**Honest residual (per no-deferral policy, documented not deferred):** with a single trusted proxy, RemoteIp trusts the client portion the proxy forwards. On Fly the strictly-better source is `Fly-Client-IP`, but that header is not `x-*` so it isn't captured by LiveView `connect_info :x_headers`; the finding explicitly accepts the XFF + trusted-proxies path. Operators must pin their Fly app/edge CIDRs into `:remote_ip_opts` `:proxies` (documented in `config/config.exs`) so the real originating client is peeled; until then SignupLive fails safe to per-connection rather than collapsing onto the proxy IP.

Full `mix precommit` green (Elixir 1.19 / OTP 28): compile `--warnings-as-errors`, format, `credo --strict`, dialyzer 0 real errors, **3273 tests, 0 failures**.

mkreyman:
Follow-up commit `d191361` fixes the CRITICAL bypass the confirmatory gate found in the FIX 3 trust boundary.

**The bug:** the previous check inferred "a real proxy hop was peeled" via `client = RemoteIp.from(headers, opts)` vs `rightmost = RemoteIp.from(headers, clients: ["0.0.0.0/0","::/0"])`, treating `client != rightmost` as trust. But `RemoteIp.type/2` checks `:clients` BEFORE its hardcoded `@reserved` list, so forcing all-clients makes `rightmost` keep a reserved suffix while `client` still skips it. An attacker-controlled `X-Forwarded-For` ending in any reserved value (e.g. `"9.9.9.9, 127.0.0.1"`) from a non-proxy public peer made `client != rightmost` → "trusted" → the attacker's `9.9.9.9` became the bucket key. That minted unlimited attacker-chosen, rotatable buckets (bypassing both the signup limit and the WebAuthn budget) and enabled victim-IP griefing. It fired under the default config and exploited `@reserved`, not `:proxies`.

**Root cause:** trust was inferred from the header (100% attacker-supplied). The only unspoofable signal is the TCP peer.

**Fix — peer-anchored** (`client_ip/1`):
- Peer ∈ configured `:remote_ip_opts[:proxies]` (CIDR-contains via `RemoteIp.Block`; Fly's `fdaa::/16` 6PN is the default) → resolve the client from the chain with plain `RemoteIp.from/2` (proxy appends the real client on the right; forged left prefixes are never returned); no usable client → per-connection.
- Peer NOT a trusted proxy (direct/public) → **ignore the header**, key on the unspoofable peer.
- No peer → per-connection.
The `client != rightmost` / forced-clients trick is removed. Shared `:remote_ip_opts` still feeds both `plug RemoteIp` and the LiveView; the default is safe because a non-proxy peer never trusts the header (proven: `fdaa::/16` never contains a public peer).

**Regression tests:** the exact `"9.9.9.9, 127.0.0.1"` attack from a public peer now keys on the peer, never `9.9.9.9` (fails against the old code); forged prepend behind a trusted proxy peel keys on the real client; genuine Fly-style trusted-proxy peer keys on the appended client; direct public peer keys on the peer; trusted-peer-no-XFF and no-peer fall to per-connection.

Full `mix precommit` green (Elixir 1.19 / OTP 28): compile `--warnings-as-errors`, format, `credo --strict`, dialyzer 0 real errors, **3275 tests, 0 failures**.

mkreyman:
Follow-up merge commit `77a422b` reconciles with master's partial fix (651a554) and replaces the connect_info IP resolution with an HTTP-pipeline, signed-session approach.

**(A) Master merge (3 signup files).** Kept this branch's implementation — a strict superset of 651a554 (attestation cap/size/verify-budget + full not-a-map hardening + action-scoped limit). master's `attestation_error` tests are preserved: a present-but-non-binary `reason` now falls through to the safe generic message, while a missing key / non-map payload stays a no-op (my web-02 cases). master's peer-IP smoke test was dropped — it asserted the removed connect_info `:peer_data` mechanism. No non-signup files were touched by master.

**(B) CRITICAL — real client, not the app IP.** The gate showed the peer-anchored code resolved to the app's own Fly IP: Fly appends the app IP as the rightmost `x-forwarded-for` entry, and RemoteIp scans right-to-left, so with `:proxies` = `fdaa::/16` only it returned the app IP → every visitor collapsed onto one bucket. Fix:
- `plug RemoteIp` uses a shared `:remote_ip_opts` with `headers: ["fly-client-ip", "x-forwarded-for"]`.
- `SignupLive.signup_session/1` (wired via the `:public_signup` live_session `session:` MFA) runs in the HTTP request, resolves the client IP, and signs it into the session; `mount/3` reads `session["client_ip"]` on both disconnected and connected mount (unforgeable — signed).
- It resolves `fly-client-ip` **directly** (single-header `RemoteIp.from`). fly-proxy overwrites that header with the real connecting client, so it's unspoofable and immune to the appended app IP. Off Fly it falls back to `conn.remote_ip` (RemoteIp-resolved from XFF), then to a per-connection bucket. All connect_info IP logic (`peer_data`/`x_headers`/`trusted_proxy_peer?`/`forwarded_client_ip`) is removed.

**(C) Mediums.** IPv4-mapped IPv6 (`::ffff:a.b.c.d`) is normalized to plain IPv4 for stable buckets. The 6PN over-broad-trust residual (a hostile same-Fly-org sibling app could set `fly-client-ip`) is documented in `config.exs` as an accepted same-org infra boundary — not a public-internet vuln.

**Tests (real Fly shapes).** `fly-client-ip` "203.0.113.7" + XFF "203.0.113.7, 66.241.125.10" → bucket `signup:ip:203.0.113.7`, never the app IP (fails against the app-IP-collapse code); two clients → two buckets; page loads never consume the bucket; a forged XFF can't override fly-client-ip; off-Fly XFF fallback; `signup_session/1` unit tests cover the nil (per-connection) and IPv4-mapped paths; master's WebAuthn input-validation tests retained.

Full `mix precommit` green (Elixir 1.19 / OTP 28): compile `--warnings-as-errors`, format, `credo --strict`, dialyzer 0 real errors, **3278 tests, 0 failures**. (One unrelated `ModelMixAnalyticsTest` fixture story-number collision flake surfaced once mid-run and passed on re-run; not caused by this change.)

mkreyman:
Follow-up commit `4394655` consolidates client-IP resolution into one shared module and applies systematic per-handler rate + size limits.

**FIX 1 — shared resolver (no more two-header collapse).** New `Loopctl.RemoteIp` resolves `fly-client-ip` and `x-forwarded-for` INDEPENDENTLY (each a singleton-`:headers` `RemoteIp.from/2`), so the wire-order collapse (Fly's app IP appended rightmost to XFF winning the right-to-left scan) can't happen. New `LoopctlWeb.Plugs.ClientIp` sets `conn.remote_ip` for the WHOLE pipeline via that resolver (replaces the two-header `plug RemoteIp`), leaving the raw peer when nothing resolves (never `0.0.0.0`). `SignupLive.signup_session/1` uses the SAME module. The merge's duplicate `LoopctlWeb.ClientIp` / `ResolveClientIp` (the latter had a latent IPv6 `MatchError` + `0.0.0.0` collapse) are deleted. `RegistrationRateLimiter` reads `conn.remote_ip` and inherits the correct client automatically. Config `:remote_ip_opts` drops the dead two-header `:headers`.

**Systematic hardening (FIX 2/FIX 3 + audit).** A SHARED per-connection `signup:actions:` budget (60/window) that every abusable handler draws from, plus a uniform size cap on every attacker-controlled string:

| handler | rate-limited | size-capped |
|---|---|---|
| validate | no (form-change) — size cap defuses echo | yes: friendly_name (120) + tenant name/slug/email (512) |
| request_attestation | yes: action budget | yes: friendly_name (120, = schema max) |
| attestation_captured | yes: action budget + WebAuthn verify budget (20) | yes: attestation_object/client_data_json/credential_id (8 KB each) |
| attestation_error | yes: action budget | yes: reason capped (500) before inspect/log |
| remove_authenticator | no (trivial state toggle) | n/a (index parsed to bounded int) |
| signup | yes: action budget + submit budget (5) | yes: tenant name/slug/email (512) |

FIX 2 (request_attestation unrated crypto + 8 MB friendly_name echo) and FIX 3 (attestation_error unrated 8 MB inspect/log) are both covered. `maybe_blank_error` is hardened against non-binary crafted values.

**Tests.** `Loopctl.RemoteIp` unit tests (fly-client-ip preferred; app-IP-collapse immunity; IPv4-mapped normalization); shared-budget throttling of request_attestation / attestation_captured / attestation_error / signup; tenant + friendly_name size caps; an 8 MB reason is capped and non-crashing.

Full `mix precommit` green (Elixir 1.19 / OTP 28): compile `--warnings-as-errors`, format, `credo --strict`, dialyzer 0 real errors (73/73 skipped, passed successfully), **3296 tests, 0 failures**.

mkreyman:
Follow-up commit `05f6301` adds a transport backstop + fixes the 3 Criticals and 2 Lows.

**BACKSTOP — 64 KB websocket frame cap** (`/live` socket, endpoint.ex). Bandit's default is 8 MB; no LiveView here needs a large inbound frame (biggest legit is the WebAuthn `attestation_captured` event, a few KB — signup/onboarding/wiki carry tiny events), so 64 KB bounds friendly_name / reason / index / tenant fields / attestation blobs at the transport, neutralizing the emoji/Zalgo/bignum/8 MB abuse class regardless of per-handler logic. Verified all 4 LiveViews (signup, tenant_onboarding, wiki_index, wiki_show) have no uploads/hooks/textareas needing bigger frames.

**FIX 1 (CRITICAL) — byte-based `truncate_to/2`.** Was grapheme-based (`String.slice`/`split_at`): multi-byte UTF-8 blew past the byte cap, and combining marks ("Zalgo") collapsed to one grapheme so the WHOLE multi-MB string was returned at O(input). Now `:binary.part(0, max)` + snap ≤ 3 bytes to a valid UTF-8 boundary — O(max_bytes). Also: `request_attestation` draws the action budget BEFORE extracting/truncating friendly_name.

**FIX 2 (CRITICAL) — bignum index.** `remove_authenticator` crashed (`SystemLimitError`, uncaught by with/else) / burned CPU on a huge numeric index. Added an O(1) `byte_size(index) <= 3` head guard before `Integer.parse`; oversized → no-op clause.

**FIX 3 (CRITICAL) — wrong-type params.** `validate` / `signup` now have `is_map(params)` head guards (non-map "tenant" → module catch-all no-op); `request_attestation` uses a binary-safe friendly_name extraction (`cap_string` returns "" for a non-binary) so `to_string` can't crash.

**FIX 4 (LOW) — dead security plug.** The `:registration_rate_limit` pipeline + `RegistrationRateLimiter` were defined but never `pipe_through`'d (they protected the removed `/tenants/register`; SignupLive now does the limiting). Deleted both.

**FIX 5 (LOW) — proxy peer as identity.** The raw-peer fallback keyed on `conn.remote_ip` even when the peer is a configured proxy (Fly's 6PN), collapsing all visitors onto one bucket. Added `Loopctl.RemoteIp.proxy?/1`; the peer fallback returns nil for a proxy peer → `signup_rate_key/2` uses the per-connection bucket.

**Tests:** emoji/Zalgo friendly_name byte-capped & non-crashing; ~2M-digit `remove_authenticator` index no-ops; non-map `validate`/`signup` and non-binary `request_attestation` friendly_name no-op; `RemoteIp.proxy?/1` CIDR membership; `signup_session/1` returns nil for a proxy peer and the peer IP for a public one.

Full `mix precommit` green (Elixir 1.19 / OTP 28): compile `--warnings-as-errors`, format, `credo --strict`, dialyzer 0 real errors (73/73 skipped, passed successfully), **3307 tests, 0 failures**.

mkreyman:
### Round 7 — completed the transport backstop (fragmented-message DoS) + validate bound

**FIX A — reassembled multi-frame cap.** `max_frame_size: 64_000` on the `/live` socket bounds a *single* frame only; Bandit reassembles fragmented multi-frame messages under `max_fragmented_message_size` (default 8 MB), so N sub-64KB continuation frames coalesce into one ~8 MB logical message and bypass the per-frame cap. The Phoenix `socket "/live", websocket: [...]` DSL rejects `:max_fragmented_message_size` (`Keyword.validate!` raises), so it is now set at the **Bandit server level** via the endpoint's http `websocket_options`:

```elixir
config :loopctl, LoopctlWeb.Endpoint,
  http: [websocket_options: [max_fragmented_message_size: 64_000]]
```

Added to `config/runtime.exs` (covers the prod release + all runtime envs), `config/dev.exs`, and `config/test.exs`. The socket-DSL `max_frame_size: 64_000` is kept — the two caps are complementary (per-frame vs reassembled-message) and coexist without a boot error (verified: `Bandit.PhoenixAdapter` forwards the endpoint http opts straight into `Bandit.child_spec/1`, whose top-level keys include `websocket_options`).

**FIX B — bound the one unrated handler.** `handle_event("validate", ...)` fires per keystroke, so rate-limiting it would harm legit typing UX. Instead it now does an **O(1) byte-size reject**: any tenant/friendly field larger than 64 KB drops the whole event as a no-op *before* any processing or echo (`byte_size/1` is O(1)). A flood of large validate frames therefore cannot pin the scheduler even if the transport cap were ever misconfigured; legit input is far smaller.

**Tests.**
- **Real transport regression** (`test/loopctl_web/endpoint_websocket_test.exs`): boots a self-contained Bandit server carrying the cap, opens a raw `:gen_tcp` WebSocket, sends a message as two sub-64KB continuation frames totalling >64KB, and asserts the connection is **CLOSED with code 1009** ("oversize fragmented message") rather than delivered.
- **Config regression**: asserts `LoopctlWeb.Endpoint` carries `http[:websocket_options][:max_fragmented_message_size] == 64_000`, so a future edit that drops it fails CI.
- **validate**: asserts a 100 KB field is dropped as a no-op and never echoed back.

**Gate:** `mix precommit` fully green — `compile --warnings-as-errors`, `format`, `credo --strict` (no issues), `dialyzer` (73/73 skipped, passed), **3310 tests, 0 failures (40 excluded)**.

Commit `cd39856`. PR remains a **draft**.

PR #358, opened 2026-07-13
Partial index on dispatches(expires_at) WHERE revoked_at IS NULL for the revoke sweep

### LINKED ISSUE #354: [Scale] Quick wins — do this week, before the epics (8 S-effort changes)
> Part of the **loopctl Scalability Remediation Roadmap** (see tracking issue). Findings verified against live code during the architecture audit.

**Do this week, before the epics.** Low-effort, high-impact — most are config or one-line changes, and several directly de-risk the "one heavy agent" repro. Each links to the epic it is a down-payment on.

1. **Add `Oban.Plugins.Lifeline`** (`rescue_after: 30m`) — S, **critical**. Stops orphaned `:executing` jobs from permanently eating queue slots. (Epic 2)
2. **Add `Oban.Plugins.Pruner`** (`max_age: ≥ 7d`) — S, high. Stops unbounded `oban_jobs` growth. (Epic 2)
3. **Partial index `dispatches(expires_at) WHERE revoked_at IS NULL`** — S. Turns the every-minute revoke sweep from a seq scan into an index range scan. (Epic 3)
4. **Move Oban queue widths into `runtime.exs` env vars** — S. Lets operators relieve a backlog without a deploy (prerequisite for Epic 4).
5. **ETS-cache tenant LLM settings** (bust on `set_llm_config`) — S. Removes a DB read + Cloak decrypt from every provider call. (Epic 5)
6. **Fail loud when `SCALE_ALERT_WEBHOOK_URL` is unset** (startup `:error` / readiness flag) — S. Ends silent alerting; a prerequisite for trusting any alert epic. (Epic 2)
7. **`Oban.insert_all` for the STH fanout** — S. Cuts N per-minute round-trips to one (down-payment on Epic 3).
8. **Set `DNS_CLUSTER_QUERY=loopctl.internal`** — S. Must land *before* any scale-out, or cross-node PubSub silently breaks (Epic 6 gate).

## Ordering note
Items 1, 2, 6 are prerequisites for Epic 2's observability to be trustworthy. Item 8 is a hard gate on Epic 6 (`fly scale count > 1`). Items 3 and 7 are safe standalone down-payments on Epic 3.

mkreyman:
## US-32.1 — Partial index on dispatches(expires_at) WHERE revoked_at IS NULL

Migration-only story (plus a behavior-parity test). Adds a partial index matching the exact predicate of the every-minute cross-tenant `RevokeExpiredDispatchesWorker` sweep, so it uses an index range scan instead of a full seq scan on the continuously-growing `dispatches` table.

### The problem
The sweep is tenant-agnostic (`WHERE revoked_at IS NULL AND expires_at < now()`, via `AdminRepo` BYPASSRLS). The only expiry index is the composite `(tenant_id, expires_at)`, whose leading `tenant_id` is absent from the predicate → Postgres cannot use it → full seq scan every 60s.

### Change
- **New migration** `20260713000000_add_dispatches_expires_at_active_index.exs`: `CREATE INDEX CONCURRENTLY IF NOT EXISTS dispatches_expires_at_active_index ON dispatches (expires_at) WHERE revoked_at IS NULL`, preceded by a self-healing `DROP INDEX CONCURRENTLY IF EXISTS`. Online build via `@disable_ddl_transaction true` + `@disable_migration_lock true` (house style, copied from `20260710130000_add_session_memories_promotion_sweep_index.exs`). Purely additive; does NOT touch RLS, ownership, or the existing composite index.
- **New test** `test/loopctl/workers/revoke_expired_dispatches_worker_test.exs`: behavior-parity guard (index usage is a planner concern). Asserts the sweep revokes expired non-revoked dispatches cross-tenant + cascades to their api_keys, leaves active ones untouched, and skips already-revoked rows.

### Acceptance criteria
- **AC-32.1.1** ✅ migration adds the partial index online (concurrently, DROP-before-CREATE, both DDL/lock attrs set).
- **AC-32.1.2** ✅ `EXPLAIN` of the worker predicate on dev DB plans `Index Scan using dispatches_expires_at_active_index` (not Seq Scan).
- **AC-32.1.3** ✅ worker code unchanged — index only.
- **AC-32.1.4** ✅ RLS/ownership untouched; existing `(tenant_id, expires_at)` index untouched.

`mix precommit` green (credo clean, dialyzer passed, 4190 tests 0 failures).

Part of #354 (item 3 of 8 — "Quick wins"). Does NOT close #354; the other 7 quick-wins remain.

### DIFF
--- a/priv/repo/migrations/20260713000000_add_dispatches_expires_at_active_index.exs
+++ b/priv/repo/migrations/20260713000000_add_dispatches_expires_at_active_index.exs
@@ -1,0 +1,56 @@
+defmodule Loopctl.Repo.Migrations.AddDispatchesExpiresAtActiveIndex do
+  use Ecto.Migration
+
+  # Epic 32 (scalability), US-32.1.
+  #
+  # `Loopctl.Workers.RevokeExpiredDispatchesWorker` runs every 60s via Oban Cron and
+  # issues a TENANT-AGNOSTIC (cross-tenant, BYPASSRLS via AdminRepo) sweep:
+  #
+  #     from(d in Dispatch,
+  #       where: is_nil(d.revoked_at) and d.expires_at < ^now,
+  #       select: %{id: d.id, api_key_id: d.api_key_id})
+  #
+  # The only expiry-related index on `dispatches` is the composite
+  # (tenant_id, expires_at) from 20260411234856_create_dispatches. Its LEADING column
+  # `tenant_id` is absent from the sweep predicate, so Postgres cannot use it → a full
+  # seq scan of the continuously-growing `dispatches` table every minute. The sweep MUST
+  # stay cross-tenant (it finds expired dispatches across all tenants), so it can never
+  # use a tenant-leading index.
+  #
+  # This partial index matches the sweep predicate exactly — the WHERE
+  # `revoked_at IS NULL` clause keeps it tiny (only live, un-revoked rows) and lets the
+  # planner serve `expires_at < now()` as an index range scan, making the sweep
+  # O(expired rows) instead of O(all dispatches).
+  #
+  # CONCURRENTLY (no table lock) since `dispatches` is a hot auth/chain-of-custody write
+  # path; the migration lock + DDL transaction are disabled because CREATE INDEX
+  # CONCURRENTLY cannot run inside a transaction. Purely additive — safe on the live prod
+  # DB. Does NOT touch RLS, ownership, or the existing (tenant_id, expires_at) index
+  # (other per-tenant lookups still use it).
+  @disable_ddl_transaction true
+  @disable_migration_lock true
+
+  @index "dispatches_expires_at_active_index"
+
+  def up do
+    # A CREATE INDEX CONCURRENTLY that fails midway (crash, lock timeout, deploy killed)
+    # leaves a permanently-INVALID index behind. On the next deploy, `IF NOT EXISTS`
+    # matches that invalid leftover BY NAME and skips the rebuild — the migration records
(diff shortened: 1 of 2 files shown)


mkreyman:
## AC-32.1.2 verification evidence (EXPLAIN)

Closing AC-32.1.2 with captured EXPLAIN evidence of the revoke sweep's exact predicate
(`WHERE revoked_at IS NULL AND expires_at < now()`) using the new partial index
`dispatches_expires_at_active_index`. This is now also enforced by a durable ExUnit
regression guard (`RevokeExpiredDispatchesWorkerTest`, describe "partial index …"),
and recorded in the migration's verification note.

Why forced (`SET LOCAL enable_seqscan = off`) for the deterministic case: the dev/test
`dispatches` table is ~empty, so the default planner correctly prefers a Seq Scan
(scanning a few pages beats an index descent). A naturally-chosen Index Scan is only
observable at scale — so the test proves the index MATCHES the predicate and is USABLE
by the planner at any table size, and the seeded capture below shows the default planner
picking it unprompted at scale.

**1. Empty table, planner forced to reveal usability (the deterministic ExUnit assertion):**

```
Index Scan using dispatches_expires_at_active_index on dispatches d0  (cost=0.12..8.14 rows=1 width=32)
  Index Cond: (expires_at < now())
```

**2. ~20k seeded rows (small selective expired set), DEFAULT planner (enable_seqscan on) — index chosen unprompted:**

```
Index Scan using dispatches_expires_at_active_index on dispatches d  (cost=0.29..16.72 rows=86 width=32)
  Index Cond: (expires_at < now())
```

No production code changed; worker behavior is untouched (AC-32.1.3 preserved).

Issue #349, opened 2026-07-11
[Scale][Epic 2] Oban resilience, retention & observability


mkreyman:
> Part of the **loopctl Scalability Remediation Roadmap** (see tracking issue). Findings verified against live code during the architecture audit.

**Priority: #2.** Contains the two cheapest criticals in the whole review (both S-effort config) plus the alerting that would have made every other incident non-silent. Highest ROI per token.

**Aggregate severity:** Critical. **Effort:** L (but front-loaded with S-effort wins).

## Core problem
Oban runs with *only* the Cron plugin. There is **no Lifeline** — an executing job orphaned by a deploy/OOM sits in `:executing` forever, permanently consuming one of the tiny (5/5/10) concurrency slots; this is the literal 17-day/110-job stall. There is **no Pruner** — `oban_jobs` grows unbounded, degrading Oban's own fetch/unique queries *and* the app's several JSONB scans over the raw table. And there is **zero Oban telemetry**: queue depth, orphan count, and discard/retry rate reach no dashboard and no alert; the health check reports HEALTHY for a queue full of orphans. Even the primary Repo's pool-checkout `queue_time` — the canonical "one heavy agent" leading indicator — is summary-only with no firing alert, and several already-emitted provider-degradation telemetry events terminate in no-ops.

## Target architecture
- Add `{Oban.Plugins.Lifeline, rescue_after: :timer.minutes(30)}` (per-queue windows for long knowledge/embedding jobs) and `{Oban.Plugins.Pruner, max_age: ≥ 7 days}` (retain past the knowledge.ex 7-day error window); confirm Stager present.
- Attach `[:oban, :job, :stop | :exception]` handlers plus a poller doing `SELECT state, count(*) FROM oban_jobs GROUP BY state` → per-queue/per-state Prometheus gauges; add an `:executing`-older-than-N-minutes orphan gauge to the `/health` degraded path.
- Add three `ScaleAlerts` signals: primary-Repo `queue_time` p95, fleet-wide discard/retry rate, and provider-error rate. Wire the dead telemetry (`[:loopctl,:llm,:blocked]`, `embedding.skipped_no_key`, `index_health.invalid`) into `metrics/0` with bounded tags.
- Harden the alert path: fail loud (startup `:error` / readiness flag) if `scale_alerts_enabled` but URL unset; retry delivery via the existing Oban webhook path with backoff; periodic re-notification while a breach persists instead of pure edge-trigger.
[...]

PR #378, opened 2026-07-13
feat(epic_34): US-34.1 Oban queue/state gauges + :executing orphan gauge; Reindexer

mkreyman:
US-34.1 (Epic 34, #349) — Oban queue/state observability.

## What
- Per-queue / per-state Oban job-count gauges (`SELECT state, count(*) FROM oban_jobs GROUP BY state`) exported to Prometheus via ScaleMetrics, bounded tags only.
- `:executing`-older-than-N-minutes **orphan gauge** — the signal that Lifeline is falling behind / a job is wedged (the 17-day/110-orphan stall was previously invisible).
- `Oban.Plugins.Reindexer` added (index bloat control); Stager handling documented.

## Why escalated to a session-level PR
The workflow implemented + review-fixed this story and pushed the branch (`d0bcce6`, incl. review-fix commit "widen dispatch_oban_stats/0 rescue, gate poller off in :test"), but its merge agent's grant didn't include `gh pr create`, so no PR was opened. The branch is clean, fully pushed, and already carries its review fixes — this PR just opens + merges the completed work. US-34.2 and US-34.3 depend on this and were skipped; they run once this lands.

## Safety
Purely additive observability + one Oban plugin (Reindexer). Bounded-cardinality metrics (state/queue only, no tenant/args tags). Poller gated off in :test. No job semantics change.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -129,6 +129,13 @@ config :loopctl,
   # `telemetry_poller` instance in LoopctlWeb.Telemetry (NOT the shared
   # tenant-label-gate poller — a review finding flagged inheriting that cadence as a
   # deviation from the AC's "bounded interval (config)" wording).
+  # - oban_metrics_poll_enabled: start the SUPERVISED oban poller child. ON by
+  #   default (purely additive in prod/dev). Forced OFF in :test (config/test.exs)
+  #   — a review finding: the poller fires at boot via `dispatch_oban_stats/0`'s
+  #   `Loopctl.MockObanStats` Mox seam, before any ExUnit test process holds the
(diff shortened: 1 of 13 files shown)


mkreyman:
Closing: US-34.1's branch is 3 behind master and conflicts heavily with US-34.4/34.5 in the scale_metrics.ex boot-path telemetry module (both renumber/recount the moduledoc metrics list). Re-implementing US-34.1 fresh on current master via the workflow (cleaner + safer than a ~300-line hand-merge of observability code) — then US-34.2/34.3 follow.

PR #396, opened 2026-07-14
Batch background embeddings into arrays (~100), map results back by index

mkreyman:
## US-37.4 — Batch background embeddings into arrays (~100), map results back by index

Background embedding previously issued one provider round-trip per record
(`ArticleEmbeddingWorker`/`MemoryEmbeddingWorker` per-item fan-out). Under bulk
ingest that is N calls for N records. This adds an array batch path that sends
`input` as a list of up to a config-driven max (~100) texts in ONE `Req.post` and
maps each returned vector back to its input **by the response `index` field**
(never by array position), cutting provider round-trips ~100x. Same model, same
per-text vector — an efficiency change, not a semantic one. The interactive
single-query path stays single-text and byte-for-byte unchanged.

### What changed
- **`EmbeddingClient.generate_embeddings/2`** (+ behaviour callback + default test
  stub): sends `input` as an array, reassembles vectors in input order from each
  entry's `index` (out-of-order `data` handled), one admission token per batch,
  empty list short-circuits with no call.
- **`Knowledge.generate_embeddings/3`**: guarded batch entry that shares the
  single-text circuit breaker, takes ONE US-37.2 concurrency slot for the whole
  batch, and records the `[:loopctl, :llm, :provider_error]` signal once per batch.
- **`Loopctl.Workers.BatchEmbeddingWorker`** (new): per-tenant drainer (unique per
  `tenant+kind`) that embeds a tenant's pending records in chunks of `batch_max`,
  issuing `ceil(N/batch_max)` calls per run. Atomic — a provider error writes no
  vectors and retries the batch as a unit. Preserves content-hash idempotency,
  BYO / permanent-error / rate-limit semantics, the US-36.2 fair-share gate, and
  tenant isolation (a batch array holds exactly one tenant's texts).
- **Enqueue sites** (article create/update/publish, bulk publish, content
  ingestion, memory `remember`) now enqueue the batch worker. Pending set is
  `embedding IS NULL`; a content change re-nulls the vector so the record
  re-enters the set (deliberately not a `hash IS NULL` predicate, which would
  loop-re-embed `update_embedding/3`'s valid nil-hash writes).
- **Migration** seeds `embedding_batch_max` = 100 (live-tunable via SystemConfig;
  in-code default applies on cache miss).
- The single-item workers are retained as the one-record embedding primitive
  (still exercised by the fair-share gate tests) and documented as superseded for
  background use.

### Acceptance criteria
- **AC-37.4.1** — array `input`, index-mapped reassembly, order preserved. Client
  test drives the real `EmbeddingClient` with deliberately out-of-order `data`.
- **AC-37.4.2** — background embedding uses the batch path; test asserts 250
  records → exactly 3 provider calls (not 250).
- **AC-37.4.3** — single-vs-batch vectors identical per index (deterministic
  stub); a provider error writes no vectors and errors/retries as a unit.
- **AC-37.4.4** — env-driven batch size, one token per batch, tenant isolation,
  no `Application.put_env` in tests.

### Quality gate
`mix precommit` green (compile clean, format, credo --strict, dialyzer, full
test — 4572 tests, 0 failures). New tests in
`test/loopctl/knowledge/embedding_client_test.exs` and
`test/loopctl/workers/batch_embedding_worker_test.exs`; three enqueue tests in
`article_embedding_worker_test.exs` updated to the batch path.

Note: the suite contains a pre-existing seed-dependent flake in
`memory_promotion_worker_test.exs` (Mox async-supervisor `$callers` race,
unrelated to this change) that passes in isolation and on most seeds.

### DIFF
--- a/lib/loopctl/knowledge.ex
+++ b/lib/loopctl/knowledge.ex
@@ -60,7 +60,7 @@ defmodule Loopctl.Knowledge do
   alias Loopctl.Llm.ProviderError
   alias Loopctl.Projects.Project
   alias Loopctl.Webhooks.EventGenerator
-  alias Loopctl.Workers.ArticleEmbeddingWorker
+  alias Loopctl.Workers.BatchEmbeddingWorker
 
   # Maximum page size for the article list endpoint (`list_articles/2`). Larger
   # limits are HONORED up to this cap so callers can paginate to exhaustion at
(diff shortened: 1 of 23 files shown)


mkreyman:
Closing the blocked first attempt. US-37.4 is being re-implemented against the amended story (AC-37.4.5 no-stranding / write-then-recall guarantee) via a fresh implement-plan run, since this branch's coalescing batch drainer stranded just-written memories embedding IS NULL. Superseded.

PR #510, opened 2026-07-23
Add put_root_layout to the browser pipeline

### LINKED ISSUE #494: Fresh self-hosted install: /signup unstyled and non-functional — browser pipeline missing put_root_layout
## Problem

On a fresh self-hosted install (master `7eeef55`, release built from the root Dockerfile), `/signup` renders **unstyled and non-functional**: the dead render starts directly with the LiveView container div — no `<!DOCTYPE>`, no `<head>`, no CSS/JS links. Visually the hero SVG icons blow up to full viewport width; functionally the WebAuthn ceremony can't run at all, because `app.js` (LiveView socket + credential hooks) is never loaded. Since signup is the only way to mint a tenant, this blocks self-hosted onboarding entirely.

## Root cause

The `:browser` pipeline in `lib/loopctl_web/router.ex` has no `put_root_layout` (and no `fetch_live_flash`):

```elixir
pipeline :browser do
  plug :accepts, ["html"]
  plug :fetch_session
  plug :protect_from_forgery
  plug :put_secure_browser_headers, %{...}
end
```

`use Phoenix.LiveView, layout: {LoopctlWeb.Layouts, :app}` (lib/loopctl_web.ex:68) sets the *app* layout only; the *root* layout (which is what carries `<head>` with the asset links — `layouts/root.html.heex` exists and is fine) is never plugged, so LiveView dead-renders bare.

## Fix (verified locally)

```elixir
plug :fetch_session
plug :fetch_live_flash
plug :put_root_layout, html: {LoopctlWeb.Layouts, :root}
plug :protect_from_forgery
```

With these two lines the page renders styled, `app.css`/`app.js` return 200, and the enrollment ceremony proceeds. Happy to PR if useful.

(If the hosted instance's signup looks fine, something environment-specific must be compensating there — but from a clean clone this reproduces deterministically.)

FinanceAlex:
Fixes #494.

On a fresh install, `/signup` renders without the root layout — no CSS bundle, no flash support, and the page is effectively non-functional. The `:browser` pipeline was missing `put_root_layout` (and `fetch_live_flash`).

**Change:** add the two standard plugs to the `:browser` pipeline. After this, `/signup` renders styled and functional.

**Testing:** running in my self-hosted deployment since 2026-07-22 (Docker, fresh install; signup page renders correctly). Compile-checked against current `master` with the Dockerfile's build image.

(One of the fixes I've been carrying locally since setting up self-hosting — sending it upstream per your suggestion. Two more follow-up PRs reference #494/#496.)

### DIFF
--- a/lib/loopctl_web/router.ex
+++ b/lib/loopctl_web/router.ex
@@ -4,6 +4,8 @@ defmodule LoopctlWeb.Router do
   pipeline :browser do
     plug :accepts, ["html"]
     plug :fetch_session
+    plug :fetch_live_flash
+    plug :put_root_layout, html: {LoopctlWeb.Layouts, :root}
     plug :protect_from_forgery
 
     plug :put_secure_browser_headers, %{


FinanceAlex:
Superseded by #523, which adds put_root_layout to the browser pipeline as part of the signup end-to-end work. Verified on our install after rebasing — fresh /signup renders styled and functional. Closing; thanks!

PR #511, opened 2026-07-23
Make WebAuthn relying-party configurable via environment

FinanceAlex:
Refs #494 (the WebAuthn hardcode noted in the issue comments).

`config.exs` hardcodes the WebAuthn relying-party (`rp_id`/`origin`) to `loopctl.com`, so the enrollment ceremony fails on any other origin — localhost or a self-hosted domain.

**Change:** allow overriding via `WEBAUTHN_RP_ID` / `WEBAUTHN_ORIGIN` env vars in `runtime.exs`. Without the env vars, behaviour is exactly as before (the block is skipped entirely), so cloud deployments are unaffected.

**Testing:** running in my self-hosted deployment since 2026-07-22 (`WEBAUTHN_RP_ID=localhost`, enrollment ceremony works). Compile-checked against current `master` with the Dockerfile's build image.

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -528,3 +528,14 @@ if config_env() == :prod do
   #
   # Check `Plug.SSL` for all available options in `force_ssl`.
 end
+
+# WebAuthn relying-party is configurable via env for non-loopctl.com deployments
+# (config.exs hardcodes loopctl.com, which breaks the enrollment ceremony on
+# localhost/self-hosted origins). Without the env var, behaviour is unchanged.
+if webauthn_rp_id = System.get_env("WEBAUTHN_RP_ID") do
+  config :loopctl, :webauthn,
+    rp_id: webauthn_rp_id,
+    rp_name: "loopctl",
+    origin: System.get_env("WEBAUTHN_ORIGIN") || "http://#{webauthn_rp_id}:4030",
+    user_verification: "preferred"
+end


FinanceAlex:
Still relevant after #523: master's signup e2e work did not include env-configurable WebAuthn rp_id/origin (checked at rebase — no WEBAUTHN_RP_ID handling in runtime.exs on master). Rebased this branch cleanly onto current master on our install; happy to update the PR if you want a different shape.

mkreyman:
Thanks for this — you found a real hole, and the diagnosis is exactly right.

`config.exs` hardcodes the relying party to `loopctl.com`, and since the WebAuthn spec requires `rp_id` to be a registrable domain suffix of the page's origin, the browser refuses the ceremony on any other domain before anything server-side is reached. What makes it worse than a broken page: enrollment is the **only** path to `human_anchored`, and that tier gates the entire chain-of-custody surface — work breakdown, dispatch, custody, token budgets. So a self-hosted deployment couldn't reach any of it. Same family as #494/#496.

Carried forward in **#545**, with your commit and authorship intact, plus a follow-up commit for three things. All three were invisible from where you were sitting, and that's on our setup, not on you: **the Elixir CI jobs don't run for fork PRs**, so this PR only ever got a GitGuardian check. You had no way to see any of it.

1. **It failed the build.** `mix loopctl.check_env_docs` (part of `mix precommit`) scans `runtime.exs` and fails on any env var missing from `deploy/FLY_SECRETS.md`. All three are documented now. There's an irony worth noting: that gate exists so self-hosters can discover env vars without reading our source — your exact audience.

2. **Placement.** The block sat at the top level of `runtime.exs`, outside the `config_env() == :prod` guard, so it applied to every environment. `dev.exs` and `test.exs` already set their own localhost relying party, so a `WEBAUTHN_RP_ID` merely *present in your shell* would silently override them — including during `mix test`, where it breaks the WebAuthn suite in a way that looks nothing like its cause. Moved inside the prod block, which is also the only environment that was actually missing an override.

3. **The origin default.** `http://#{rp_id}:4030` is loopctl's *dev* port over *http*. WebAuthn only runs in a secure context — localhost is the one `http://` exception — so for a real domain that origin can never work, and it fails in the browser rather than at configuration time, which is much harder to debug. It defaults to `https://#{rp_id}` now, and `WEBAUTHN_ORIGIN` is documented as required for localhost or a non-standard port. Your own setup with `WEBAUTHN_RP_ID=localhost` will need `WEBAUTHN_ORIGIN=http://localhost:<port>` set explicitly — worth checking against what you're running.

Closing this in favour of #545. Only because our API token can't update a fork branch that pulls in a `.github/workflows/` change (missing `workflow` scope) — nothing to do with the contribution. Your commit is in that branch as-is.

mkreyman:
Shipped as #545 — merged just now (commit on master). Your commit is in it with your authorship, and the changelog credits you.

Two things the review caught that are worth knowing, since they affect your own deployment:

1. **A blank `WEBAUTHN_RP_ID` used to be treated as set**, which would have replaced the working relying party with `rp_id: ""`. Now trimmed and treated as unset, with host-shape validation.
2. **`WEBAUTHN_RP_ID` is effectively write-once.** Changing it on a deployment that already has enrolled authenticators invalidates all of them via `rpIdHash` — which takes break-glass and every reauth-gated custody operation with it. Now documented in `deploy/FLY_SECRETS.md`. Worth knowing before you change yours.

And the practical one for you: set **`WEBAUTHN_ORIGIN` explicitly** (e.g. `http://localhost:4000`). The origin now derives from `PHX_HOST` rather than from `rp_id`, so a localhost or non-standard-port setup needs it stated.

Thanks — this was a genuine gap, and it blocked the whole human-anchor path for anyone self-hosting.

Issue #496, opened 2026-07-22
Signup hard-depends on Fly.io secrets adapter — tenant creation impossible on self-hosted installs (+ LiveView crash instead of error)


FinanceAlex:
## Problem

Third fresh-self-host blocker in the signup path (after the root-layout and WebAuthn-origin ones in #494): completing signup **hard-depends on Fly.io**.

`Loopctl.Secrets` has exactly two adapters: `FlyAdapter` (production, Fly GraphQL API) and the Mox test stub. Off Fly, `FlyAdapter` returns `{:error, :fly_not_configured}` (`lib/loopctl/secrets/fly_adapter.ex:41`), which surfaces as `{:error, {:audit_key_storage_failed, :fly_not_configured}}` from `Tenants` (`lib/loopctl/tenants.ex:283`) — and `SignupLive.complete_signup/3` has **no case clause** for it, so the LiveView process crashes (`lib/loopctl_web/live/signup_live.ex:481`, `CaseClauseError`) and the user just sees the button do nothing. Tenant creation is impossible on any non-Fly deployment.

Log evidence:

```
FlyAdapter: FLY_APP_NAME or FLY_API_TOKEN not configured
** (CaseClauseError) no case clause matching: {:error, {:audit_key_storage_failed, :fly_not_configured}}
    lib/loopctl_web/live/signup_live.ex:481: LoopctlWeb.SignupLive.complete_signup/3
```

## Suggestion

1. A non-Fly secrets adapter for self-hosted installs. We're running a minimal file-backed one (JSON, base64 values, 0600, atomic tmp+rename, selected via `SECRETS_ADAPTER=local_file` in runtime.exs) — verified: signup completes. Happy to PR it (with the threat-model caveat that it's for trusted-single-machine installs only).
2. Independently: `complete_signup/3` should handle `{:error, {:audit_key_storage_failed, _}}` gracefully (error banner) instead of crashing the LiveView — that turns any storage failure into silent button-does-nothing.

mkreyman:
Fixed in #523 (merged c58ea0d): local file secrets adapter (SECRETS_ADAPTER=local_file) + graceful audit-key-storage error handling in the signup LiveView.

PR #512, opened 2026-07-23
Add opt-in file-backed Secrets adapter for non-Fly deployments

FinanceAlex:
Refs #496.

Tenant signup hard-depends on Fly secrets storage: off-platform, `FlyAdapter` fails signup with `{:audit_key_storage_failed, :fly_not_configured}` — a fresh non-Fly install cannot create its first tenant.

**Change:** an opt-in `Loopctl.Secrets.LocalFileAdapter` (same behaviour contract as `FlyAdapter`; atomic write via temp+rename; file mode 0600; path from `SECRETS_FILE`). It is selected in `runtime.exs` **only** when `SECRETS_ADAPTER=local_file` is set — default (Fly) behaviour is unchanged, so this is purely additive for cloud deployments.

I'm aware this one touches product direction (self-hosting) more than the other two PRs, so treat it as a proposal — happy to rework the shape (adapter name, env contract, or moving it under the epic-39 private-tier umbrella) to fit your vision.

**Testing:** this adapter has been running in my self-hosted deployment since 2026-07-22 (signup, key issuance, daily use). Compile-checked against current `master` with the Dockerfile's build image.

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -528,3 +528,10 @@ if config_env() == :prod do
   #
   # Check `Plug.SSL` for all available options in `force_ssl`.
 end
+
+# Tenant secrets: opt-in file-backed adapter for non-Fly deployments
+# (FlyAdapter hard-fails signup off-platform: {:audit_key_storage_failed, :fly_not_configured}).
+# Default behavior (Fly) is unchanged unless SECRETS_ADAPTER=local_file is set.
+if System.get_env("SECRETS_ADAPTER") == "local_file" do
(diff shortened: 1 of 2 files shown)


FinanceAlex:
Superseded by the LocalFileAdapter that shipped with #523 — and it is a better version of the same idea (atomic rename, fsync, 0600 before secret bytes). The env interface (SECRETS_ADAPTER=local_file, SECRETS_FILE) matched our deployment exactly, so the switch was zero-effort and our existing secrets file worked as-is. Closing; thanks!

PR #534, opened 2026-07-30
Run scale-nightly on beelink, and fix CI granting on the wrong database

mkreyman:
## Why

`scale-nightly` was the largest GitHub-hosted line item in this repo — **~143 min of billed runner time per night** across its 14 files, roughly **4,300 min/month**, more than every other hosted job here combined.

## Isolation is load-bearing, so it had to be replaced, not dropped

Each file seeds an ~80k prod-floor corpus and cleans up only its own tenant. A leaked corpus poisons the next file's planner statistics and flips `assert_hnsw_index` into a false RED. The hosted job got that isolation from a **fresh Postgres service container per matrix job**.

beelink runs one long-lived Postgres, so isolation now comes from a **per-matrix-file database**, derived from the file name. That is equivalent for the property these tests assert: planner statistics live in the per-database `pg_statistic`, so two databases in one cluster cannot see each other's `ANALYZE` results any more than two containers could. The cluster does share CPU and buffers — but that affects *timing*, not the plan **choice** being asserted, and the matrix runs serially on the single runner anyway.

## Consequential details

- **`services:` removed.** beelink already publishes Postgres on 5432, so a container mapping `5432:5432` cannot bind and the job dies at `Initialize containers` before its first step. The host instance has pgvector 0.8.2, which is all the pgvector image was supplying.
- **Step order changed, and the order matters.** The container pre-created its database via `POSTGRES_DB`, so the old job could grant *before* `ecto.create`. A per-file database does not exist until `ecto.create` makes it, so it is now create → grant → migrate. `ALTER DEFAULT PRIVILEGES` stays *between* create and migrate because it only affects tables created after it runs.
- **A drop step with `if: always()` is new and mandatory.** The disposable container made cleanup free; on a long-lived server, 14 corpora at ~850 MB would add ~12 GB **per night** until the disk filled. `WITH (FORCE)` terminates lingering pool connections that otherwise make `DROP` fail and silently strand the space.

## Separately: this fixes a real defect in the `test` job

`test` sets `MIX_TEST_PARTITION: ci`, so `config/test.exs` points all three repos at `loopctl_testci` — but the psql steps were hardcoded to `loopctl_test`.

On a hosted runner that was merely useless. On the self-hosted box `loopctl_test` is the **developer's local test database**, so every CI run issued `GRANT` and `ALTER DEFAULT PRIVILEGES` against it while the database the suite actually used never received them.

Measured before this change: `loopctl_testci` had `USAGE` but **not** `CREATE` on schema public and one default-ACL entry, while the local `loopctl_test` had two. The suite passed regardless, which is exactly what kept it invisible.

Also drops `DATABASE_URL` from the test steps: `config/runtime.exs` reads it only under `config_env() == :prod`, so in `MIX_ENV=test` it was dead config that *looked* like it selected the database.

## Verification

Ran the full workflow step sequence for one scale file on beelink: database created (54 chars), grants applied, migrations run, gate green, database dropped, zero leftovers.

That run also **caught a defect in my own name derivation**, since fixed: piping `basename` into `tr` feeds tr the trailing newline, which is non-alphanumeric and became a trailing underscore in the database name. Bash parameter expansion is used instead. All 14 derived names are unique and at most 54 bytes, inside Postgres's 63-byte identifier limit.

The runbook guard (`scale_verification_runbook_test.exs`, which parses `strategy.matrix.scale_file`) still passes — 9 tests, 0 failures. Full local gate green: 6396 tests, 0 failures.

The matrix itself is `schedule || workflow_dispatch`, so this PR's own checks do not exercise it; a `workflow_dispatch` run against this branch does that, and its result is reported in a comment below before merge.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -773,8 +773,10 @@ jobs:
 
       # The disposable service container used to make cleanup free. On a long-lived
       # Postgres it is mandatory: each of these databases holds an ~80k-article
-      # prod-floor corpus with its vectors and HNSW index, measured at roughly 850 MB, so
-      # 14 of them left behind would add ~12 GB to the box EVERY NIGHT until it filled.
+      # prod-floor corpus with its vectors and HNSW index. Observed live during the first
+      # beelink run: 1345 MB for remediation_matrix alone (they vary widely — the
+      # calibration file needs only ~14 MB). A full matrix left un-dropped is therefore
+      # on the order of 10 GB added to the box EVERY NIGHT until it filled.
       #
       # `if: always()` so a failed or cancelled gate still reclaims its space — an
       # abandoned corpus is exactly what a timed-out 40-minute seed leaves behind.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -145,30 +145,51 @@ jobs:
               END IF;
             END \$\$;
           "
-          PGPASSWORD=postgres psql -h localhost -U postgres -d loopctl_test -c "
-            GRANT ALL PRIVILEGES ON ALL TABLES IN SCHEMA public TO loopctl_app;
-            GRANT ALL PRIVILEGES ON ALL SEQUENCES IN SCHEMA public TO loopctl_app;
+
+      # These steps target "loopctl_test$MIX_TEST_PARTITION", NOT a hardcoded
+      # loopctl_test, and that is a FIX rather than a tidy-up.
+      #
+      # MIX_TEST_PARTITION is `ci` for this job, so config/test.exs points all three repos
+      # at loopctl_testci while these psql steps were still hardcoded to loopctl_test.
+      # On the hosted runner that was merely useless (the container's only database was
+      # loopctl_test). On the self-hosted box loopctl_test is the DEVELOPER'S local test
+      # database, so every CI run was issuing GRANT and ALTER DEFAULT PRIVILEGES against
+      # it while the database the suite actually used never received them.
+      #
+      # Observed before this change: loopctl_testci had USAGE but not CREATE on schema
+      # public and one default-ACL entry, while the local loopctl_test had two — CI's
+      # grants, landed on the wrong database. The suite passed anyway, which is what made
+      # this invisible: nothing in it currently needs loopctl_app to hold CREATE.
+      #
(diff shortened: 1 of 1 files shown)


mkreyman:
## Validation complete — 14/14 green on beelink

`workflow_dispatch` run [30518981195](https://github.com/mkreyman/loopctl/actions/runs/30518981195) against this branch: **conclusion `success`, all 14 matrix jobs `success`**, no non-success job.

| | Hosted (previous nightly) | beelink |
|---|---|---|
| Compute | ~143 min **billed** | **71.5 min**, free |
| Wall clock | ~40 min (parallel) | 75.8 min (serial, one runner) |
| Longest job | 40.0 min (streaming_export) | 19.4 min |

Roughly **half the compute time** — 4 schedulers and local disk against hosted's 2 cores. Wall clock is longer because the 14 jobs serialise on the single `beelink-loopctl` runner, which is irrelevant for an 02:00 cron.

**Cleanup verified:** `select count(*) from pg_database where datname like 'loopctl_test_sn_%'` returned **0** afterwards, and the host's free space was unchanged at 769 G.

**The `test` job fix is confirmed by direct observation.** Before, `loopctl_testci` had `CREATE=f` on schema public with 1 default-ACL entry, while the developer's local `loopctl_test` carried 2 — CI's grants, landed on the wrong database. After: `loopctl_testci` reads `CREATE=t` with 2 entries.

**One correction pushed since validation.** The cleanup comment cited ~850 MB per corpus, an estimate carried from an unrelated pre-existing database. The live run measured **1345 MB** for `remediation_matrix`, and showed the spread is far wider than one number suggests — the calibration file needs ~14 MB. The comment now cites the observed figure and the spread. Comment-only; the validated behaviour is unchanged.

**Unrelated flakiness seen while doing this**, filed here so it is not mistaken for fallout: `EmbeddingsSideTableReadsTest` (recurrence after #487/#508/#519, reproduces with `--seed 68823`) and `RetrievalEvalTest` "a drop inside tolerance is not a regression" both fail intermittently in full-suite runs and pass in isolation. Neither is touched by this PR.

mkreyman:
Reopening to force an Actions run: the push of 983fce6 updated the PR head and GitGuardian reacted, but no workflow run was created for the synchronize event.

PR #587, opened 2026-08-05
Tune autovacuum where the tables actually accumulate

### LINKED ISSUE #579: Autoanalyze has never run on the four largest tables: n_live_tup was wrong by 3 orders of magnitude
## Symptom

Autoanalyze had **never run** on any of the four largest tables. `pg_stat_user_tables` reported row counts wrong by three orders of magnitude:

| table | `n_live_tup` reported | actual rows |
|---|---|---|
| `articles` | **358** | **85,053** |
| `article_links` | **6,026** | **1,417,624** |
| `article_embeddings` | **344** | 84,171 |
| `article_access_events` | **4,863** | 56,124 |

`autoanalyze_count = 0` and `last_autoanalyze IS NULL` on all four. Only `oban_jobs` was being analyzed (27 times).

I ran `ANALYZE` manually on 2026-08-04, so the numbers above are corrected **now** — but nothing prevents them drifting back.

## Why it happened (not a bug — a threshold)

Autoanalyze fires at `autovacuum_analyze_threshold (50) + autovacuum_analyze_scale_factor (0.1) * reltuples`.

For `articles` that is `50 + 0.1 * 85,053` ≈ **8,555 changed rows**. The corpus was bulk-loaded during the 2026-07-30 consolidation and has grown by fewer than that since, so the trigger legitimately never fired. `oban_jobs` churns constantly, which is why only it was analyzed.

So this is a **tuning gap**, not a broken autovacuum. `SHOW autovacuum` is `on` and no table sets `autovacuum_enabled = false`.

## Why it matters

1. **Planner accuracy.** `reltuples` in `pg_class` happened to be roughly right (the restore populated it), so plans were not catastrophically misled — but column statistics (histograms, `n_distinct`, MCVs) are what actually drive selectivity, and those age with the data. On a corpus that grows steadily this drifts silently.
2. **Anything reading `n_live_tup` was simply wrong** — capacity checks, dashboards, any "how big is the corpus" report. A monitor reading 358 for an 85,053-row table is worse than no monitor.

## Fix

Per-table autovacuum tuning, following the pattern in the `rgn-performance` playbook — pick the knob that matches the access pattern:

- **Append-heavy / low-update** (`articles`, `article_links`, `article_embeddings`, `article_access_events`, `audit_log` partitions) — tune the **insert-driven** trigger:
  ```sql
  ALTER TABLE <table> SET (autovacuum_analyze_scale_factor = 0.05,
                           autovacuum_vacuum_insert_scale_factor = 0.1);
  ```
- Do **not** use the update-churn `vacuum_scale_factor` knob on these; they accumulate rather than churn (`n_dead_tup` ≈ 0, which is correct and not a missing vacuum).

Gotchas from the playbook that apply here:

- A partitioned **parent** rejects storage autovacuum params — set them on **partitions only**, and bake the `ALTER` into whatever creates new `audit_log_*` partitions (`AuditPartitionWorker`), since a `PARTITION OF` child does not inherit reloptions.
- Make the stamp **self-healing**: apply when the reloption is missing, guarded so it never needlessly re-`ALTER`s (the `ALTER` takes `SHARE UPDATE EXCLUSIVE` and would fight the autovacuum it is meant to enable).
- A table rebuild silently drops reloptions — re-apply after any copy-swap.

This is a catalog-only change (instant, no rewrite), so it is safe to ship as a normal migration rather than out-of-band DDL.

## Acceptance

- Each named table carries the reloptions (`SELECT relname, reloptions FROM pg_class WHERE ...`).
- `last_autoanalyze` becomes non-NULL and recent on the hot tables within a normal growth cycle.
- New `audit_log_*` partitions are created already tuned.
- A test asserts the partition-creation path stamps the reloptions, so this cannot silently regress.

mkreyman:
Closes #579.

Autoanalyze had **never run** on the four largest tables — `autoanalyze_count = 0`, `last_autoanalyze IS NULL` — so `pg_stat_user_tables` reported row counts wrong by three orders of magnitude:

| table | reported `n_live_tup` | actual |
|---|---|---|
| `articles` | 358 | **85,053** |
| `article_links` | 6,026 | **1,417,624** |

Anything reading `n_live_tup` — a capacity check, a dashboard — was simply wrong.

## Not a broken autovacuum — a threshold gap

`autovacuum` is `on`, no table disables it, and `oban_jobs` has been analyzed 27 times. The trigger is `50 + 0.1 * reltuples` ≈ **8,555 changes** for `articles`; the corpus was bulk-loaded at the 2026-07-30 consolidation and has grown by less than that since, so it legitimately never fired. `oban_jobs` churns constantly, which is why only it was covered.

## The knob follows the access pattern

These tables **accumulate** rather than churn — `n_dead_tup` measured ~0 on `article_links` and `article_access_events`, 475 on `articles` — so the **insert-driven** trigger is the right one. The update-churn `vacuum_scale_factor` would wait on dead tuples that never arrive (which is also why "no autovacuum, only autoanalyze" on an append-only table is correct, not a symptom).

Halving the analyze scale factor takes `articles` from ~8,555 changes to ~4,303, and `article_access_events` to ~2,868. Catalog-only and instant, so it rides a normal migration.

## Partitions

The `audit_log` **parent** is deliberately excluded — a partitioned parent has no heap and rejects storage autovacuum params outright. Partitions are stamped individually: existing ones by the migration, new ones by `AuditPartitionWorker`, because `CREATE TABLE ... PARTITION OF` does not inherit reloptions.

The stamp is **self-healing** and fires on the already-exists branch too, covering a partition created before this code and one created untuned during a deploy window (a table rebuild also silently drops reloptions). It's guarded on the reloption being *absent* rather than applied unconditionally — `ALTER TABLE` takes a `SHARE UPDATE EXCLUSIVE` lock and would fight the very autovacuum it's enabling. A tuning failure logs and never fails partition creation.

## Two things I got wrong, both caught by tests

- A `DO $$` block takes **no bind parameters** — the first version raised at runtime. It's now a parameterised lookup plus an `ALTER` that interpolates only a name generated from a year/month integer pair.
- The self-healing test first failed for the *wrong reason*: I reset reloptions through `AdminRepo` while the worker writes through `Repo`, and those hold **separate sandbox transactions** — so the test couldn't observe a repair that actually happened.

**Both guards mutation-verified:** making the stamp a no-op fails the self-healing test, and the coverage test asserts its row set is non-empty first so it cannot pass vacuously.

`mix precommit` green: 6694 tests, 0 failures.

### DIFF
--- a/lib/loopctl/workers/audit_partition_worker.ex
+++ b/lib/loopctl/workers/audit_partition_worker.ex
@@ -22,6 +22,14 @@ defmodule Loopctl.Workers.AuditPartitionWorker do
   @default_retention_days 90
   @future_months 3
 
+  # Autovacuum tuning stamped onto every partition (#579). Append-only leaves accumulate
+  # rather than churn, so the INSERT-driven trigger is the knob — the update-churn
+  # `vacuum_scale_factor` would wait on dead tuples that never arrive. Kept in lockstep with
+  # `20260804230000_tune_autovacuum_on_append_heavy_tables.exs`, which stamps the partitions
+  # that already existed; a partitioned PARENT has no heap and rejects these params entirely.
+  @analyze_scale_factor "0.05"
+  @vacuum_insert_scale_factor "0.1"
+
   @impl Oban.Worker
   def perform(_job) do
     create_future_partitions()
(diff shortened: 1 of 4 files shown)


mkreyman:
**Correction to the round-2 rationale, before this merges.**

The review routed all DDL in this worker through `AdminRepo`, justified as repairing a live production outage — that `DATABASE_URL` is the grants-only `loopctl_app`, that migrations run as owner `loopctl_admin`, and so every stamp had been failing into the fail-soft warning in prod.

That's the *documented* architecture, not the deployed one. Measured on prod today:

- `DATABASE_URL` connects as role **`loopctl`**
- `ADMIN_DATABASE_URL` is **unset**, so `config/runtime.exs:247` falls it back to the same URL — both repos are the same role
- the `audit_log` partitions are **owned by `loopctl`**

So `Repo` did own them and the DDL was succeeding. The claim also has a testable consequence that didn't hold: if `Repo` lacked ownership, `CREATE TABLE ... PARTITION OF` would have been failing too, and partitions for m05–m11 exist.

**Keeping the change** — it's right for a different reason. DDL requires *ownership*, and AdminRepo is the repo whose role owns these relations (migrations run through it, so it created every partition). Neutral against the deployed single-role setup; necessary against the documented split. `e0e599c` rewrites the comment to say that, and to say explicitly not to simplify it back to `Repo`: the moment the split is actually deployed, every statement starts failing into the fail-soft warning and the tuning silently stops — precisely the undetected shape #579 was.

Correcting the record rather than the code, because a comment asserting a false fact about production is worse than no comment.

Issue #588, opened 2026-08-05
SystemConfig cache is primed AFTER the Endpoint starts, so vector reads silently fall back to the retired legacy embedding column on every boot


mkreyman:
Found while investigating #578 (why the retired 657 MB `articles_embedding_hnsw_idx` still shows 26 index scans). It is not a stray consumer — it is this.

## The defect

`Loopctl.SystemConfig.get_int/2` returns its **in-code default on a `:persistent_term` cache miss** (`lib/loopctl/system_config.ex:54-61`):

```elixir
def get_int(key, default) when is_binary(key) and is_integer(default) do
  case :persistent_term.get(pt_key(key), :__miss__) do
    value when is_integer(value) -> value
    _ -> default
  end
rescue
  _ -> default
end
```

For `embedding_side_table_reads` the default is **`0`** — which means *read the legacy `articles.embedding` column*.

And the cache is primed **after the web server is already accepting requests**:

- `LoopctlWeb.Endpoint` is the **last child** in the supervision tree (`lib/loopctl/application.ex:159`)
- `Loopctl.SystemConfig.refresh()` is called **after `Supervisor.start_link(children, opts)` returns** (`application.ex:~172`)

So on every boot there is a window in which the Endpoint (and Oban) serve work while the flag reads `0`. Every vector request landing in it takes the **legacy** branch — silently reverting a cutover that completed on 2026-07-22.

## Evidence this is actually happening in production

`articles_embedding_hnsw_idx` (legacy) has **26** scans against **1,695** on `article_embeddings_hnsw_dim_1536_idx` (live) — ~1.5%. That ratio is the shape of a brief per-boot window multiplied over ~5 days of Fly deploys and machine restarts. A genuinely un-gated legacy path would show a continuous, proportional share; a code search found none — every `ORDER BY <=> $const` path against `articles.embedding` is behind this flag.

The three affected paths, all flag-gated:
- `search_semantic_legacy/6` (`lib/loopctl/knowledge.ex:6967`) — every semantic/hybrid search request
- suggest-links / `ProposalGate` / `ArticleLinkingWorker` (`knowledge.ex:4906`, `knowledge/vector_search.ex:465-472`) — every `knowledge_create` and ingest
- `novelty_distance_legacy_query/4` (`knowledge.ex:6224`) — the dedup gate

## It is worse than a boot race

[...]

PR #590, opened 2026-08-05
#578 Retire the legacy 657 MB articles_embedding_hnsw_idx

### LINKED ISSUE #578: Retire the legacy 657MB articles_embedding_hnsw_idx: 26 scans vs 1695 on its replacement, and it evicts the live index from cache
## Symptom

`articles_embedding_hnsw_idx` — the **pre-US-41.1 legacy vector index** — is **657 MB with 26 scans**, while the live side-table index it was replaced by has **1,695 scans**.

It is **87% of the entire `articles` index footprint** (753 MB of indexes over a 136 MB heap — a 5.5x index-to-heap ratio) and it serves a read path that was cut over on 2026-07-22.

```
 indexrelname                        | idx_scan | size
 articles_embedding_hnsw_idx         |       26 | 657 MB    <- legacy
 articles_search_vector_idx          |     1856 | 59 MB
 articles_tenant_title_active_idx    |       28 | 10184 kB
 ...
 article_embeddings_hnsw_dim_1536_idx|     1695 | 658 MB    <- live (other table)
```

## Why it matters beyond disk: it is evicting the live index from cache

`shared_buffers` is **1536 MB**. There are **two ~657 MB HNSW indexes** competing for it, only one of which is live.

Measured on prod, a cold vector search:

```
 calls | mean_ms | shared_blks_hit | shared_blks_read | blk_read_time
     1 |    8044 |            6989 |            17915 |       7926 ms
```

**7,926 ms of the 8,044 ms was disk I/O — 98.5%.** The same query warm: **0 ms**.

That cold read happened after a Postgres restart, so it is not a steady-state figure — but it is exactly what users get after any restart, failover, or eviction, and the eviction pressure is caused by carrying a redundant 657 MB index. Dropping it frees the space *and* roughly doubles the live index's chance of staying resident.

## Corroboration from our own instrumentation

The #551 retirement observer ran on 2026-08-04 and independently recorded the same picture:

```
 observed_on | side_table_reads | legacy_columns_present | legacy_index_scans
 2026-08-04  | 1                | {articles,memories}    | {"articles_embedding_hnsw_idx": 26,
                                                           "memories_embedding_idx": 0,
                                                           "memories_live_embedding_hnsw_idx": 0}
```

`side_table_reads = 1` — the side table is the live path. Only **one** observation row exists so far, so retirement is not yet "owed" by that mechanism's own rule.

## Open questions to settle before dropping

1. **What are the 26 scans?** Non-zero means *something* still reads it. Identify the caller before dropping, or the drop turns a slow path into a broken one.
2. **Stats window.** Counters date from the 2026-07-30 consolidation, so 26 scans is over ~5 days, not all time.
3. **`memories` has the same shape** — `memories_embedding_idx` and `memories_live_embedding_hnsw_idx` both at 0 scans. Same decision, likely same change.
4. Dropping the **index** is separable from and far safer than dropping the **column**. Do the index first; the column is a schema migration and belongs to the #551 retirement flow.

## Fix

1. Identify the 26 scans' source.
2. `DROP INDEX CONCURRENTLY articles_embedding_hnsw_idx;` out-of-band, codified as a guarded `DROP INDEX IF EXISTS` migration (`CONCURRENTLY` cannot run inside the migration transaction).
3. Re-measure the cold-read figure afterwards and record before/after.

## Acceptance

- Cold vector-search `blk_read_time` materially lower after the drop (re-measure the same statement via `pg_stat_statements`).
- No regression in semantic search / ANN recall — the retrieval-eval CI job stays green.
- The #551 observer keeps reporting `side_table_reads = 1` with the legacy index gone from its scan map.

mkreyman:
## What

Retires the pre-US-41.1 HNSW index on the legacy `articles.embedding` column
(`articles_embedding_hnsw_idx`) — but only on installs whose reads have been cut over to
the `article_embeddings` side table.

Measured on the hosted instance 2026-08-04 (`pg_stat_user_indexes` / `pg_stat_statements`):

| index | size | scans |
|---|---|---|
| `articles_embedding_hnsw_idx` (legacy column) | 657 MB | 26 |
| `article_embeddings_hnsw_dim_1536_idx` (live read path) | 658 MB | 1,695 |

`shared_buffers` is 1536 MB, so the two ~657 MB indexes evict each other. A cold vector
search measured 8,044 ms of which 7,926 ms (98.5%) was `blk_read_time`; the same statement
warm was 0 ms.

## The one real design decision — the drop is GUARDED on the read flag

`embedding_side_table_reads` defaults to `0` **in code**, and no migration seeds the row.
So on a fresh self-hosted install — and in the test DB — the shipped read path is still the
legacy `articles.embedding` column. An unconditional drop would leave that shipped-default
path with no index at all: a seq scan + top-N sort over the whole corpus, which under
`HeavyRead`'s per-read `SET LOCAL statement_timeout` surfaces as
`{:error, :heavy_read_overloaded}`. That is the #588 failure re-introduced through
configuration instead of ordering.

So `up/0` reads `system_configs.embedding_side_table_reads` **directly from the row** (not
`SystemConfig.get_int/2` — the `:persistent_term` cache answers its in-code default on a
miss, and a migrator process is exactly where that cache may be cold) and drops only when it
is `1`. Prod is already `1`, so this is a real drop there; a fresh install keeps the index
that serves its own read path.

**The alternative — seeding `embedding_side_table_reads = 1` in the migration — was
implemented and measured, then rejected.** Flipping the flag in the test DB puts the entire
suite on the side-table path and produces **34 failing behavioural tests** (semantic search
pool/starvation edges, combined-priors ordering, article-linking, backfill, and the read-path
test that asserts the in-code default). That is a read-path cutover decision with its own
blast radius and the known side-table ANN flake, not an index retirement. For comparison, an
unconditional drop with the flag left alone produces 3 failures plus a hard break of the
scale harness (`ScaleSeed.seed/2` refuses to seed without an HNSW index on `articles`,
`PlanAssertions.assert_hnsw_index/1` asserts the `articles` scan IS that index) — so
"unconditional drop, pin the tests" is the expensive option too, and it degrades the
shipped default. The guard gets the prod win with neither cost.

Known wart, recorded rather than deferred: an install that flips the flag to `1` **after**
this migration has run keeps the index, because a migration does not re-run. That is not
silent — `Loopctl.Embeddings.LegacyRetirement` discovers legacy indexes BY COLUMN, so the
leftover reappears in its scan map and its evidence/deadline verdict names it. Rerunning the
`DROP INDEX CONCURRENTLY` out-of-band is the remedy, and it is in the CHANGELOG.

## Scope kept tight

The `articles.embedding` **column is untouched** — still dual-written, still the
backfill/reconciliation source and the documented revert target. The `memories` indexes are
untouched (see open question 3).

This change touches no role or auth gate, so nothing in the CLAUDE.md chain-of-custody
checklist changes. No new environment variables, so `mix loopctl.check_env_docs` is not in
play.

## Open questions from the issue — answered

1. **What are the 26 scans?** The GH #588 boot window: the `SystemConfig` cache was primed
   after the Endpoint started, so every boot had an interval in which
   `embedding_side_table_reads` read its in-code default `0` and vector reads used the legacy
   column. #588 is merged (a398ace, PR #589) — `Loopctl.SystemConfig.CachePrimer` is now a
   supervised one-shot child ordered after `AdminRepo` and before `Oban`/`Endpoint`. A code
   search found no un-gated legacy ANN path. Confirm the counter has stopped advancing across
   deploys before running the out-of-band drop (below).
2. **Stats window.** The counters date from the 2026-07-30 consolidation, so 26 scans is
   roughly five days, not all time. That is consistent with the boot-window explanation
   (a handful of scans per deploy) and inconsistent with a live consumer.
3. **`memories` has the same shape — same decision?** No, and deliberately not.
   `memories_embedding_idx` and `memories_live_embedding_hnsw_idx` are a documented keep-both
   decision (US-38.4 / TC-38.4.3, `test/loopctl/memory/dual_index_recall_test.exs`): the FULL
   index serves `recall(include_superseded: true)` and the PARTIAL one serves default live
   recall. Both sit on the legacy `memories.embedding` column that the default (flag `0`)
   read path still uses, so their 0 prod scans reflect prod's flag, not deadness. #578's own
   Fix and Acceptance sections name only the articles index. Not dropped here.
4. **Index before column?** Yes — this PR drops only the INDEX. The column is separable and
   far riskier: it is still dual-written and is the backfill/reconciliation source.

## Migration shape

`@disable_ddl_transaction true` + `@disable_migration_lock true` (both mandatory —
`CONCURRENTLY` is illegal inside a transaction block and Ecto's migration advisory lock is
itself transaction-scoped), explicit `up/0`/`down/0`, raw `execute/1` because Ecto's
`create index(..., concurrently: true)` helper emits no `IF NOT EXISTS`.

`down/0` is unconditional (a rollback that skipped would leave a cut-over install without
the index it is rolling back TO) and leads with a DROP that fires **only when the index is
INVALID** — the leftover a failed `CREATE INDEX CONCURRENTLY` produces, which
`IF NOT EXISTS` would otherwise treat as present forever (wiki `ed00911b`). An unconditional
drop-then-create would destroy and rebuild a good 657 MB index on every rollback. The
`WITH (m, ef_construction)` clause is single-sourced from `Loopctl.Repo.HnswIndex`, never
hard-coded. The migration comment records that the rollback rebuild at prod scale needs
`maintenance_work_mem` well above the 64 MB default or it silently falls back to the slow
on-disk build (wiki `753fbf69`).

The interaction with `20260410022906_add_embedding_hnsw_index.exs` (capability-based create)
and `20260624120000_reconcile_hnsw_index_name.exs` (rename to the canonical name) is
unchanged and correct: on a fresh `mix ecto.reset` both still run, then this migration
decides. No already-applied migration was edited.

## Tests

- **`test/loopctl/repo/drop_legacy_articles_embedding_hnsw_index_migration_test.exs`** (new)
  drives the SHIPPED `up/0`/`down/0` through `Ecto.Migration.Runner.run/8` inside
  `Sandbox.unboxed_run(AdminRepo, ...)` (CONCURRENTLY is illegal in the sandbox transaction),
  `async: false` with the documented global-catalog-state rationale, `on_exit` restoring the
  canonical state. Covers: drop when the flag is `1`; KEEP when the row is absent (the shipped
  default) and when it is explicitly `0` (an operator revert); idempotence in both directions;
  the live side-table index and the legacy COLUMN left intact; the INVALID-leftover rebuild;
  and that a VALID index is NOT needlessly rebuilt (`relfilenode` unchanged).
- **TC-20.2.7** (`knowledge_embedding_test.exs`) no longer asserts a bare "legacy index
  exists". It asserts the invariant that survives the retirement: the live per-dimension
  side-table index is present AND `indisvalid` (`pg_indexes` also lists INVALID indexes, and a
  row count proves nothing — wiki `753fbf69`), and the legacy index is present exactly when
  the legacy column is the live read path. A migration that dropped unconditionally fails it.
- **Acceptance bullet 3** — two new `LegacyRetirementTest` cases cover an index
  DISAPPEARING between observations, the shape this drop produces and one no existing test
  had: it must read as neither a scan nor a counter that went BACKWARDS (an absence read as a
  value would jam the trigger at `:not_due` forever on exactly the installs that completed the
  retirement), and it must not make the scan check vacuous for indexes that survived it.

## Verification

- `mix precommit` green — 6725 tests, 0 failures (compile `--warnings-as-errors`, format,
  `credo --strict`, dialyzer).
- The two `:scale` gates CI runs on every PR, run the way CI runs them on an isolated
  `MIX_TEST_PARTITION` database: `scale_seed_test.exs` (5 tests, 0 failures — including
  `hnsw_index_present?/0`) and `embedding_dimension_plan_scale_test.exs` with
  `EMBEDDING_PLAN_GATE_DIMENSIONS=1536` (8 tests, 0 failures).
- **Acceptance bullet 2** — `mix loopctl.retrieval.eval --mode both --fail-on-regression`
  run locally: `status: OK (no regression)`, no losers.
- The `:scale_nightly` matrix is SKIPPED on pull requests (it has no cron and only a
  `workflow_dispatch` arm), so the one nightly file whose assertions actually depend on this
  index was run locally the way that job runs it — its own `MIX_TEST_PARTITION` database,
  `mix ecto.create` + `mix ecto.migrate`, then `SCALE_TESTS=true SCALE_NIGHTLY=true mix test
  --only scale_nightly test/loopctl/knowledge/distant_pairs_novelty_scale_test.exs`:
  **3 tests, 0 failures** in 332 s over a fresh ~80k-row corpus. That file `EXPLAIN`s
  `Knowledge.novelty_distance_query/4`, which dispatches on the read flag and therefore
  explains the LEGACY query in the test env, and it asserts `PlanAssertions.refute_seq_scan/1`.
  It passing on a database migrated FROM SCRATCH with this migration included is direct
  evidence that the guard keeps the legacy plan index-served. The remaining `:scale_nightly`
  files are unaffected for the same reason, structurally: with the flag row absent the
  migration is a catalog no-op, so `articles_embedding_hnsw_idx` is still present and
  `indisvalid` in the test database after it runs (verified by query), leaving every scale
  corpus identical to master's.
- The read-flag guard is mutation-verified, not merely covered: replacing
  `if side_table_reads_live?() do` with `if true do` (an unconditional drop) fails the two KEEP
  cases. The guard is live code, not decoration.
- End-to-end migration behaviour verified on a fresh partition database in BOTH directions:
  with no flag row the index is retained and the skip is logged; with the row set to `1` the
  migration emits `DROP INDEX CONCURRENTLY IF EXISTS articles_embedding_hnsw_idx` and
  `pg_indexes` returns 0 hnsw rows for `articles`.

### Still owed by the operator (cannot be done from CI)

Acceptance bullets 1 and 3 of the Fix section are production measurements and need a
`fly ssh console`:

1. Confirm `pg_stat_user_indexes.idx_scan` for `articles_embedding_hnsw_idx` has stopped
   advancing across deploys since a398ace shipped.
2. `DROP INDEX CONCURRENTLY articles_embedding_hnsw_idx;` out-of-band (the migration is then
   a no-op there, by design).
3. Re-measure the same statement in `pg_stat_statements` and record before/after
   `blk_read_time`. The post-drop plan check must be `EXPLAIN`-based and show
   `Index Scan using article_embeddings_hnsw_dim_1536_idx` — a row count proves nothing.

Closes #578

### DIFF
--- a/.claude/skills/knowledge-wiki/SKILL.md
+++ b/.claude/skills/knowledge-wiki/SKILL.md
@@ -160,11 +160,16 @@ number, is the load-bearing invariant. To monitor and keep it healthy over time:
   `blk_read_time`. `articles.embedding` is still dual-WRITTEN — it stays the backfill/reconciliation
   source, and the column was NOT dropped — but on an install where the drop has run, setting the flag
   back to `0` puts reads on a column with no ANN index: a seq scan + top-N sort over the corpus, which
-  under `HeavyRead`'s per-read `SET LOCAL statement_timeout` surfaces as
-  `{:error, :heavy_read_overloaded}`. A revert is therefore an INCIDENT action, not a routine toggle —
-  rebuild the index first (`down/0` of `20260805120000_drop_legacy_articles_embedding_hnsw_index.exs`,
-  and raise `maintenance_work_mem` well above the 64 MB default or the ~657 MB build silently falls back
-  to the slow on-disk path, wiki `753fbf69`). The drop migration is GUARDED on that same flag read
+  trips `HeavyRead`'s per-read `SET LOCAL statement_timeout`. That cancel is a raised `Postgrex.Error`
+  (57014) rendering `504 db_statement_timeout` — NOT `{:error, :heavy_read_overloaded}` (only the
+  `TenantGate` concurrency shed produces that tuple) and so NOT the labelled keyword degrade, which
+  matches the shed alone. Semantic search returns no results at all. A revert is therefore an INCIDENT
+  action, not a routine toggle — rebuild the index FIRST (an explicit `CREATE INDEX CONCURRENTLY`, or
+  `down/0` of `20260805120000_drop_legacy_articles_embedding_hnsw_index.exs`; raise
+  `maintenance_work_mem` well above the 64 MB default or the ~657 MB build silently falls back
+  to the slow on-disk path, wiki `753fbf69`) — and if you rebuilt via a rollback, flip the flag to `0`
+  BEFORE the next migrate runs, since the rollback makes that migration PENDING again and the next
+  deploy would re-drop the index you just rebuilt. The drop migration is GUARDED on that same flag read
   straight from `system_configs`, so an install still reading the legacy column keeps its index and a
   fresh self-hosted install is unaffected. `Loopctl.Embeddings.LegacyRetirement` discovers legacy
   indexes BY COLUMN, so an install that cuts over AFTER the migration ran still gets the leftover named
--- a/.claude/skills/knowledge-wiki/SKILL.md
+++ b/.claude/skills/knowledge-wiki/SKILL.md
@@ -176,7 +176,9 @@ number, is the load-bearing invariant. To monitor and keep it healthy over time:
   in its scan map. The executable surface enforces this: `mix loopctl.embeddings revert` REFUSES while
   the legacy index is absent (capability-detected, `--force` to override) and `mix loopctl.embeddings
   status` reports its presence. Operator procedure:
-  `docs/runbooks/embedding-dimension-cutover.md` (Reverting).
+  `docs/runbooks/embedding-dimension-cutover.md` — Retiring the legacy articles ANN index (the
+  drop, the baseline above, and the `pg_stat_statements` query for the AFTER reading) and
+  Reverting.
 - **Bulk (re)embed / backfill is a live-DB hazard.** Unthrottled it 504s the live wiki — per-row HNSW
   index maintenance saturates the small Fly Postgres and starves concurrent heavy-read searches past
   their `statement_timeout`. Throttled id-range keyset pattern: wiki `7a4187fd`.
--- a/.claude/skills/knowledge-wiki/SKILL.md
+++ b/.claude/skills/knowledge-wiki/SKILL.md
@@ -173,7 +173,10 @@ number, is the load-bearing invariant. To monitor and keep it healthy over time:
   straight from `system_configs`, so an install still reading the legacy column keeps its index and a
   fresh self-hosted install is unaffected. `Loopctl.Embeddings.LegacyRetirement` discovers legacy
   indexes BY COLUMN, so an install that cuts over AFTER the migration ran still gets the leftover named
-  in its scan map.
+  in its scan map. The executable surface enforces this: `mix loopctl.embeddings revert` REFUSES while
+  the legacy index is absent (capability-detected, `--force` to override) and `mix loopctl.embeddings
+  status` reports its presence. Operator procedure:
+  `docs/runbooks/embedding-dimension-cutover.md` (Reverting).
 - **Bulk (re)embed / backfill is a live-DB hazard.** Unthrottled it 504s the live wiki — per-row HNSW
   index maintenance saturates the small Fly Postgres and starves concurrent heavy-read searches past
   their `statement_timeout`. Throttled id-range keyset pattern: wiki `7a4187fd`.
--- a/.claude/skills/knowledge-wiki/SKILL.md
+++ b/.claude/skills/knowledge-wiki/SKILL.md
@@ -173,9 +173,12 @@ number, is the load-bearing invariant. To monitor and keep it healthy over time:
   straight from `system_configs`, so an install still reading the legacy column keeps its index and a
   fresh self-hosted install is unaffected. `Loopctl.Embeddings.LegacyRetirement` discovers legacy
   indexes BY COLUMN, so an install that cuts over AFTER the migration ran still gets the leftover named
-  in its scan map. The executable surface enforces this: `mix loopctl.embeddings revert` REFUSES while
-  the legacy index is absent (capability-detected, `--force` to override) and `mix loopctl.embeddings
-  status` reports its presence. Operator procedure:
+  in its scan map. `mix loopctl.embeddings revert` REFUSES while the legacy index is absent
+  (capability-detected, `--force` to override) and `mix loopctl.embeddings status` reports its
+  presence — but that refusal binds only where `mix` exists (source checkout / self-hosted from
+  source). **A release ships no `mix`**, so on the hosted instance a revert goes through
+  `bin/loopctl eval` or plain SQL and the guard never runs; there the runbook's rebuild-first
+  ordering is the only thing standing between you and a tenant-wide semantic-search outage. Operator procedure:
   `docs/runbooks/embedding-dimension-cutover.md` — Retiring the legacy articles ANN index (the
   drop, the baseline above, and the `pg_stat_statements` query for the AFTER reading) and
   Reverting.
--- a/.claude/skills/knowledge-wiki/SKILL.md
+++ b/.claude/skills/knowledge-wiki/SKILL.md
@@ -153,6 +153,22 @@ number, is the load-bearing invariant. To monitor and keep it healthy over time:
   NEVER write the flag — it is `:persistent_term`-cached VM-globally, so a write leaks across the whole
   node. `Loopctl.ConfigEmbeddingReadPathTest` fails the build on a second writer and on any
   `:embedding_read_path` config outside `config/test.exs`.
+- **Reverting the flag is now an UNINDEXED read path on a cut-over install (GH #578).** The legacy
+  `articles_embedding_hnsw_idx` (657 MB / 26 scans on prod 2026-08-04, vs 658 MB / 1,695 scans for the
+  live side-table index, `pg_stat_user_indexes`) is retired: with `shared_buffers` at 1536 MB the two
+  ~657 MB indexes evicted each other, and a cold vector search measured 8,044 ms of which 7,926 ms was
+  `blk_read_time`. `articles.embedding` is still dual-WRITTEN — it stays the backfill/reconciliation
+  source, and the column was NOT dropped — but on an install where the drop has run, setting the flag
+  back to `0` puts reads on a column with no ANN index: a seq scan + top-N sort over the corpus, which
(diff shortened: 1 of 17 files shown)


mkreyman:
Independent verification against the **live hosted instance**, since this is the one property CI cannot check — the test DB has no `system_configs` row for the flag, so it can only ever exercise the skip path unless a test seeds one (this PR's does, correctly).

```
column_name | data_type
value       | bigint

key                        | value | pg_typeof
embedding_side_table_reads | 1     | bigint
```

`value` is `bigint`, and Postgrex decodes it to an Elixir integer, so `rows` is `[[1]]` and `match?([[1]], rows)` matches. **The guard is live on prod and the drop will fire there** — it is not silently inert, which is the failure mode a `match?` against an assumed column type usually has.

The other direction holds too: a fresh install has no row, `rows` is `[]`, no match, index kept.

Two notes on the flag-guard itself, which is a better design than the issue asked for. Guarding on the read flag rather than dropping unconditionally is the right call, and the reasoning in the migration comment is correct: an unconditional drop would leave the SHIPPED-DEFAULT read path with no index at all, which is #588's failure re-introduced through configuration instead of ordering. Worth keeping that comment intact.

One figure is stale rather than wrong: the CHANGELOG and the migration header quote 1,695 scans on the live index. Re-measured a few hours later it is **1,863**, while the legacy index is still at **26** — zero movement, with no deploy in that window. That delta strengthens the claim the header makes (it is the observation that separates "boot-window artifact" from "steady un-gated reader"), so it is worth refreshing if the fix rounds touch these files anyway. Both indexes measure 658 MB; "657 MB" is a rounding of the same number, not a second reading.

mkreyman:
**The acceptance bullet I said could not be checked before merge has now partially checked out.** Correcting my own caveat on #578 in the direction of more confidence.

#588 merged at 04:39 UTC; loopctl **v440** deployed and the machine restarted at **04:44:30 UTC** (`flyctl machine status 48e13eebd96e08`). So the app has now booted once WITH `CachePrimer` in the tree, and has served an hour of live traffic on that build.

Reading just taken:

```
article_embeddings_hnsw_dim_1536_idx | 1877   (+14 in the last ~20 min, +182 over 6h)
articles_embedding_hnsw_idx          |   26   (unmoved across all of it)
```

The legacy index took **zero scans across a post-fix boot**. Before #588, each boot contributed to that 26. That is the mechanism-level confirmation the third acceptance bullet was asking for, at n=1 deploy — not conclusive on its own, and the counter is destroyed by this migration so it cannot be accumulated further, but it is a real post-fix observation and it points the right way rather than the wrong one.

What this changes for the merge: the residual risk I named on #578 was "the drop merges before the boot-window fix is empirically confirmed." It is no longer unconfirmed — it is confirmed once. The remaining unknown is unchanged and unchangeable by waiting: that no future code change reintroduces a legacy read path. `LegacyRetirement` discovering by COLUMN rather than by name is what covers that, and this PR keeps it.

Issue #605, opened 2026-08-05
Consolidation auto-apply: decide class ownership, and gate on reversibility instead of human calibration


mkreyman:
Replaces #594, which was written around a two-week **human approve/reject calibration** stage. That premise is dead: there will be no human in the review loop, so a stage whose gate is accumulated human verdicts can never open. Closing it and restating the goal against the real constraint.

## Where consolidation actually stands

Stage 1 (#584, PR #593) shipped: the nightly pass emits evidence-carrying proposals for four defect classes and writes nothing to `articles` / `article_links` / `conflict_resolutions`. Read surface is `GET /api/v1/knowledge/consolidation`. `review_status` is `:pending | :approved | :rejected` with **no apply path anywhere**.

## The inconsistency to resolve first

**The class consolidation was going to auto-apply first is already being auto-applied — somewhere else.**

`:contradiction_candidate` is the only class whose apply writes zero article rows and has a genuine idempotent undo (upsert a `conflict_resolutions` row). PR #601 now does exactly that from the **nightly lint worker**: auto-judging auto-generated `potential_conflict` pairs as `classification: :redundant, disposition: :dismiss`, capped at 2000/night against a 500/night promoter so the queue converges.

So consolidation still *reports* a class that another writer is already *resolving*. Two writers on the same pairs is the failure mode to avoid — the same "second scheduler over one corpus" #584 named. **Decide the ownership before building any apply path:** either consolidation stops proposing that class, or the lint worker's judge moves into consolidation and the lint worker calls it.

## What replaces the calibration gate

Calibration existed to answer "is this class safe to auto-apply?" using human verdicts. Without humans, the same question has to be answered from the system's own record. Concretely, a class earns auto-apply when **its own outcomes are measurable and its reversal is cheap** — not when someone signs off.

That reduces to two properties, both checkable in code rather than by accumulating approvals:

1. **The apply writes nothing irreversible.** This is the hard gate, and it is stricter than it looks — see below.
2. **The proposal's evidence is verifiable at apply time**, not just at propose time. The corpus moves between the nightly scan and the write.

## The constraint that reorders everything: archive is TERMINAL for articles

`Article`'s `@valid_transitions` has no `{:archived, _}` and there is no unarchive function. The only escape is a `user+` PATCH with a caller-supplied status — i.e. a human, which is what we do not have. `bulk_ops.ex`'s docstring calling archive "reversible" is wrong for articles, and `knowledge_archive`'s tool description says the same thing.

[...]

PR #607, opened 2026-08-05
Let the duplicate class apply itself, gated on agreement and reversibility

mkreyman:
Second half of #605, stacked on #606. **The first time this pass writes to `articles`.**

## What it does

Keeps the richest article of a confirmed duplicate group and **unpublishes** the rest. That is the whole action, bounded at 25 groups per nightly run.

## Three restrictions, each load-bearing

**1. Unpublish, never archive.** `:archived` is terminal for an article — no `{:archived, _}` transition, no unarchive function, so the only way back is a `user+` PATCH. `{:published, :draft}` and `{:draft, :published}` are both real. **A class earns auto-apply by being reversible, not by being confident.**

**2. Two consecutive runs must agree.** This is what replaces the human approve/reject stage. A proposal applies only if the same fingerprint appeared in the *previous* report too — so anything transient (a duplicate created and cleaned up between runs, a title edited mid-scan, a half-finished import) is gone by the next night and never acted on.

It costs one night of latency and removes the entire class of *"acted on a state that was already resolving itself"* — which is most of what a reviewer actually catches.

Deliberately **not** a confidence score. Confidence is the machine grading its own homework; agreement across two independent observations of a moving corpus is evidence.

**3. The winner is recomputed at apply time.** The corpus moves between the scan and the write. Every member is re-checked as still published; a group with fewer than two live members has resolved itself and is **skipped, not forced** — otherwise the apply would unpublish the last remaining copy.

## Mutation-verified — and the first attempt caught a bad test of mine

Removing the two-run filter **did not fail the suite.** My single-run test was passing through the "fewer than two reports" short-circuit and never exercised the fingerprint match at all — a test that couldn't fail, guarding the most important property here.

Added a test where two reports exist but the duplicate appears only in the newest. With that in place:

- removing the two-run filter → fails
- removing the apply-time re-verification → fails the skip test

Fail-soft in the worker, like `consolidate/2`: a raise here must not abort a pass whose other steps already succeeded. A failure costs one night of duplicate cleanup, the cheapest thing in the run to lose.

`mix precommit` green — **6886 tests, 0 failures**.

Not applied autonomously, and stated in the code: `:generic_title` needs a generated replacement and can collide on the active-title index (potentially *manufacturing* a duplicate proposal next night); `:stale_entry` is a time threshold with no correctness signal behind it, so auto-archiving on age would silently delete the corpus over time.

### DIFF
--- a/lib/loopctl/knowledge/bulk_ops.ex
+++ b/lib/loopctl/knowledge/bulk_ops.ex
@@ -27,9 +27,16 @@ defmodule Loopctl.Knowledge.BulkOps do
 
   ## Operations
 
-  - `archive/3` — `:draft`/`:published` → `:archived` (reversible; links left
-    intact but inert in published-only graph traversal, AC-27.12.3).
-  - `unpublish/3` — `:published` → `:draft` (reversible).
+  - `archive/3` — `:draft`/`:published` → `:archived`. **NOT reversible in code.** The row
+    is retained and links are left intact but inert in published-only graph traversal
+    (AC-27.12.3), so nothing is destroyed — but `Article`'s `@valid_transitions` has no
+    `{:archived, _}` entry and there is no unarchive function, so the ONLY way back is a
+    `user+` PATCH carrying an explicit status. An earlier version of this line called it
+    "reversible"; that was wrong for articles and is exactly the sentence an unattended
+    caller would rely on before archiving something it cannot restore.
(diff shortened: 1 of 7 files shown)


mkreyman:
Superseded by a clean re-application on master — this branch was stacked on #606 and the squash-merge diverged its history. Same diff, same tests, reopened as the branch below.

PR #674, opened 2026-08-12
Retry the asset toolchain fetch instead of failing a deploy on a CDN hiccup

mkreyman:
## The fault

Twice in fifteen minutes today the Lint job died at `mix assets.setup`:

```
** (RuntimeError) Couldn't fetch .../tailwindcss-linux-x64: {true, {:error, :socket_closed_remotely}}
```

Both times `format`, `credo`, `dialyzer`, `test`, `security`, the retrieval eval and the scale
tests had all passed. The fetch is the first thing in that job to touch the network, and Lint
gates `Deploy` — so a GitHub CDN hiccup **skipped the deploy and left master red** for a reason
with nothing to do with the change under test. #670 and #671 sat merged-but-undeployed until I
re-ran it.

## Why it recurs rather than being random

`_build` is cached and the binaries live under it — but **`actions/cache` only SAVES on a key
miss.** A cache populated by a run where this step had not yet succeeded is never updated, so
every later run restores it *without* the binaries and downloads them again. The fetch happens
on every Lint run, and is exposed every time.

## The fix

Three attempts with linear backoff, in CI **and in the Dockerfile**.

The Dockerfile half is the same defect where it costs more: `assets.deploy` runs
`tailwind loopctl`, which auto-installs the binary when missing, so the image build reaches
the same CDN and a failure there fails a **deploy** rather than a check. It hasn't fired yet —
it's one line of the identical fault, so it's fixed here rather than left to be discovered
during an outage. Splitting the install out is what makes the retry possible and leaves
`assets.deploy` a pure build step.

**A retry, not a wider cache key.** The fetch is idempotent and `assets.setup` is a no-op once
the binaries exist, so three attempts cost nothing on the happy path; changing the cache key
would rebuild every dependency to fix a download.

## Verification

Verified **in both directions** under POSIX `sh` (what a Dockerfile `RUN` uses, not bash):

- a stub that fails twice then succeeds → recovers on attempt 3;
- a stub that always fails → prints the exhaustion message and **exits 1**, rather than
  falling through to the success path.

That second one is the case that matters: a retry wrapper that swallows a persistent failure
would turn a broken build green, which is worse than the flake it replaces.

`mix precommit` green. YAML validated.

Reviewed inline (this session's system prompt forbids dispatching review agents; per CLAUDE.md
the gate is satisfied by the best available reviewer with that stated).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -124,8 +124,28 @@ jobs:
       # That gap stopped being tolerable once /enroll shipped a ceremony that is
       # entirely client-side: for that page, broken JS is a broken feature with a
       # green build. esbuild and tailwind are standalone binaries — no Node.js.
+      #
+      # RETRIED, because this step reaches out to the network on EVERY run and gates the
+      # deploy. `_build` is cached, but `actions/cache` only SAVES on a key miss, so a cache
+      # populated by a run where this step had not yet succeeded is never updated — every
+      # later run restores it without the binaries and downloads them again. Measured
+      # 2026-08-12: two runs fifteen minutes apart both died here with
+      # `Couldn't fetch .../tailwindcss-linux-x64: {:error, :socket_closed_remotely}`, after
+      # format and credo had already passed, which skipped Deploy and left master red for a
+      # GitHub CDN hiccup that has nothing to do with the change under test.
+      #
+      # A retry, not a wider cache key: the fetch is idempotent and `mix assets.setup` is a
+      # no-op once the binaries exist, so three attempts cost nothing on the happy path,
+      # while changing the cache key would rebuild every dep to fix a download.
       - name: Install asset toolchain
-        run: mix assets.setup
+        run: |
+          for attempt in 1 2 3; do
+            if mix assets.setup; then exit 0; fi
+            echo "assets.setup failed (attempt $attempt/3); retrying in $((attempt * 5))s"
+            sleep $((attempt * 5))
+          done
+          echo "assets.setup failed after 3 attempts"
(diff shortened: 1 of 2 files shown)


mkreyman:
## Correction: the retry alone is not the fix

Pushed a second commit, because CI showed the first one was the wrong shape.

The retry ran **exactly as written** — three attempts, linear backoff, clean `exit 1`. It was still not enough, and the log says why:

```
attempt 1/3 → socket_closed_remotely
attempt 2/3 → HTTP/1.1 503 Service Unavailable
attempt 3/3 → socket_closed_remotely
assets.setup failed after 3 attempts
```

That is not a hiccup, it's an outage — ~30s of backoff cannot outlast it. **Three Lint runs in half an hour** died this way, each after format, credo, dialyzer and the full suite had passed.

So the dedicated cache is the fix and the retry is only smoothing:

- `_build` was already cached, but `actions/cache` **only saves on a key miss** — a cache first populated by a run where this step hadn't yet succeeded never gains the binaries, so every later run restores it without them and re-fetches. That's why it recurred rather than being random.
- The key hashes `config/config.exs` rather than naming the versions. Both versions live there, and a hand-copied key drifts **silently**: `tailwind.install --if-missing` sees a binary present and won't re-fetch, so a stale key would build against the wrong toolchain and say nothing. Hashing can only cost an unnecessary re-fetch — the harmless direction.

Do not lengthen the backoff instead: the cache removes the upstream dependency, the retry only covers the brief miss.

(The CDN has since recovered — verified HTTP 200 — so the blocked PRs can proceed independently of this.)

PR #751, opened 2026-08-25
Replay the exact-scan GUCs in the recall diagnostic, so its EXPLAIN describes the read it is explaining

mkreyman:
Replay the exact-scan GUCs in the recall diagnostic, so its EXPLAIN describes the read it is explaining

VectorRecallDiagnostics wraps every assertion that ever produced the left: [] signature and,
on an empty result, prints the plan of the query that just came back empty. Its whole value
rests on one property, which its own comment states: an EXPLAIN at session defaults can
describe a different execution than the read it is diagnosing, so the wrapper re-applies the
GUCs HeavyRead set.

The previous commit added enable_indexscan and enable_bitmapscan at that same chokepoint and
did not extend the list, so the diagnostic reopened exactly the divergence it exists to close.
It would have printed an HNSW index plan for a read that could not have used the index -
worse than printing nothing, because the plan looks authoritative and sends the reader after
the ANN.

Both branches need it, and the session-defaults branch is the one that is easy to miss. Its
existing comment is right that a bare HeavyRead.all/2 carries no hnsw_* opts and therefore
emits no ANN SET LOCAL - but no ANN GUC was never no GUC at all. Exact-scan forcing is applied
to EVERY heavy read when the flag is on, bare reads included, so that branch has to arm it too.

force_exact_scan?/0 becomes public for this, with a docstring saying why. The diagnostic reads
the SAME source of truth HeavyRead reads, which is the opposite of the rule the chokepoint
guard in hnsw_dead_entry_recall_test.exs follows, and the comment says so: a guard needs an
INDEPENDENT source or it moves with the thing it guards and goes vacuous, while a diagnostic
needs the SAME source or it stops describing the read. Both mistakes were made in this branch
already, one in each direction, which is why the distinction is written down next to the code.

Found by the enhanced review on the parent branch, where it was the one confirmed finding that
fell outside the reviewed diff.

### DIFF
--- a/.claude/skills/tenancy-rls/SKILL.md
+++ b/.claude/skills/tenancy-rls/SKILL.md
@@ -57,7 +57,7 @@ reads through `Loopctl.HeavyRead` (`lib/loopctl/heavy_read.ex`), which owns the 
    (`heavy_read.ex:1520-1538`) RAISES unless EVERY base-table source — `from`, every join, and every
    subquery, recursively — carries a conjunctive `x.tenant_id == ^tenant_id` bound to the passed
    `tenant_id`; `test/loopctl/heavy_read_guard_test.exs` additionally bars direct `HeavyReadRepo` calls.
-   Agent-memory reads need a SECOND predicate: `all_memory/4` (`:1090`) also requires a conjunctive
+   Agent-memory reads need a SECOND predicate: `all_memory/4` (`:1187`) also requires a conjunctive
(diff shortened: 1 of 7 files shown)


mkreyman:
Superseded by a branch cut from the post-merge master. This branch was cut from the pre-squash feature branch, so after #749 squash-merged it conflicted and CI never ran on it. Same commit, reopened as a clean PR.

PR #759, opened 2026-08-26
Stop deleting the compiled tree at the start of every CI job

mkreyman:
Stop deleting the compiled tree at the start of every CI job

Ported from home_care_billing PR 1369, where this took compile across a whole run from
638 seconds to under 10 and wall clock from 9m44s to 5m22s.

Two facts, and acting on either alone makes things worse rather than better.

First, actions/checkout defaults to clean true, which runs git clean with -ffdx. The -x
deletes ignored files: _build, deps, and the Dialyzer PLT. On a persistent self-hosted
runner every job began by throwing away the compiled tree the previous job had left at
that same path.

Second, the _build entry in the actions/cache steps was there to compensate and mostly
could not. Mix records the working directory it compiled at and force-recompiles the
whole project when it does not match, writing it in the compile manifest at
lib/mix/lib/mix/compilers/elixir.ex line 962 and reading it at line 104. A GitHub cache
is shared across runners and a repo sits at as many absolute paths as there are runners
on the host, so a restored cache avoided a rebuild only when it happened to have been
saved by a job at the same path. That was measured at roughly two jobs in six.

Verified on the other repo that removing the cache alone makes it worse and stays worse
on a second pass, because checkout keeps wiping the tree so no warmth can accrue.

This repo keeps its PLT in priv/plts rather than in _build, so the clean excepts that
path too. Getting that wrong would cost a full PLT rebuild on every run. The narrow
tailwind and esbuild binary cache under _build is deliberately left alone: those are
downloaded binaries, not a compiled tree.

Full write-up, including two remedies that look right and are not, check_cwd false and
bind-mounting the runner work directory: loopctl KB article 09245ceb.


mkreyman:
Review gate, stated honestly: the adversarial reviewer dispatched for this branch went idle twice without emitting a report, so this diff did not get its own independent audit. What it did get instead:

- The same change was audited adversarially in three sibling repos (home_care_billing #1369 by a 15-agent pipeline, and cron_books and ecommerce-friendly by independent reviewers). Every finding from those was carried over and applied here: git clean patterns anchored, the PLT exception narrowed from the whole priv/plts directory to the two artifact filenames, the asset toolchain binaries scrubbed rather than inherited across the branch boundary, and the header note corrected — it claimed deleting priv/plts costs a full PLT rebuild, when the dialyxir core plts actually live in each runner's own MIX_HOME and were never in the clean's path.
- The two repo-specific facts that made those findings land differently elsewhere were checked here directly rather than assumed: this repo pins Elixir 1.19.5, so the cwd-manifest rationale in the header is true here (it is false for ecommerce-friendly, which pins 1.18.4 and needed a different explanation); and the PLT cache key already carries runner.name, which cron_books' key did not.
- The base was verified fresh before merging. It was 9 commits behind and master has been merged in. cron_books' port of this change was cut from a stale base and its whole diff turned out to be an artifact, which is why this is now checked rather than assumed.

Measured against the master baseline (run 32922925658): Test 239s -> 135s, Lint 87s -> 39s, Dialyzer 53s -> 31s, Retrieval Eval 47s -> 22s.

Issue #803, opened 2026-09-12
Webhook intake, triage trio, Postgres claiming, DurableServer stage machine


mkreyman:
Part of the agent delivery loop epic, #800. Design: https://github.com/mkreyman/loopctl-runner/blob/main/docs/agent-delivery-loop.md §2, §3, §4, §6.

Depends on the contract shape from #801 and the predicate interface from #802. Do not start this
before those two have settled their interfaces.

### Scope

1. **Webhook intake** (§2) — a GitHub webhook, not a poller. Correlate the GH issue with the
   existing HomeCareBilling support row. The delivery id is the idempotency key.
2. **Triage trio orchestration** (§4) — dispatch `business-analyst`, `elixir-systems-architect`
   and the senior-engineer lens **in one message** so they run concurrently. They create AND
   review the story in ONE pass. Disagreement does not trigger another round; it escalates.
3. **Claiming** (§6):
   `UPDATE stories SET claimed_by = $1, claimed_at = now(), state = 'claimed' WHERE id = $2 AND claimed_by IS NULL RETURNING id`
   A Postgres transaction. NOT `Group`, NOT the existing channel-claim path — see below.
4. **Stage machine** (§3) as one `DurableServer` per story, keyed `story/<id>`, state in object
   storage or EKV. It must survive a loopctl deploy and a runner crash.

### Why claiming is a Postgres transaction

The existing `channel_claim` is **not exclusive on this fleet**: every session authenticates as
one `agent_id`, so a second claimant gets the existing claim back idempotently as though it had
won. Two sessions have already built the same issue in parallel because of it (KB
`b447b16b-05ea-452e-a5a4-c747fec950c3`, `0264d7e0-7e5a-4ce2-b848-90edd6cef23b`). A convergent
registry has the same property and is also wrong here.

### Acceptance

- Concurrent claimants: exactly one wins, proved by a test.
- A DurableServer survives a simulated control-plane restart mid-story and resumes at the right
  stage.
- A replayed webhook delivery creates no second story.

mkreyman:
## Correction: "merge on passed reviews plus pipeline" is circular as written

Verified on 2026-09-12: `required_approving_review_count` is **0** on both `master` and
`staging` of `home_care_billing`. Branch protection requires six CI contexts and **no human
approval at all**.

So "reviews passed" in the design can only mean *our own agent review gate* — which makes the
merge path an agent-written change, reviewed by agents, merged with no human anywhere. The
[...]

PR #823, opened 2026-09-13
Merge precondition: both delivery gates run again over the real pull request (#803)

mkreyman:
The second run of both delivery gates, over the real pull request (#803, design
`agent-delivery-loop.md` §5 "Both gates run twice" and §9).

The triage run judges the trio's **predicted** touch list, and nothing binds an
implementing session to that prediction — so a gate that ran only there certifies a guess.
This is the run that gates a merge.

## What has to be true to merge

All four, and every one can only **add** a refusal:

1. **Gate A** over the trio's outputs — `:proceed`, on a unanimous `story` verdict.
2. **Gate B** at `:merge` — `:clear`, or `:prove_effect` cleared by an effect proof the
   call carried. A failing proof routes to Gate A; a `:prove_effect` with no proof is a
   refusal, never a pass.
3. **The design's hard bound** — 12 files, 1,000 changed lines — applied **on top of** the
   configured limits, so a trigger document may tighten it and may not loosen it, and read
   from the forge's own diffstat rather than the length of a file list it may have
   truncated.
4. **Custody** — `verified_status: :verified`, set by a verifier dispatch whose lineage is
   separate from the implementer's.

## Shape

`Loopctl.Delivery.MergePrecondition` is split by what it touches:

- **`judge/1` is pure.** Every fact is an argument and any of them may arrive as
  `{:error, reason}` — which is exactly how an unreachable GitHub, a project with no
  repository and a diff that did not parse present themselves. That is what makes every
  fail-closed path decidable with no I/O at all.
- **`evaluate/3`** gathers the facts and calls `judge/1`. Writes nothing.
- **`enforce/3`** adds the one write a refusal implies.

## The custody clause reuses the L4 comparison rather than restating it

`Loopctl.Progress.merge_custody_status/1` delegates the separation to
`verify_recorded_separation/2` — the same clause `verify` runs — so a change to what
"separate lineage" means moves both at once.

Where it deliberately differs is **missing provenance**. `verify` may legitimately reach a
story with no verifier dispatch (`request-review` is optional) and its caller-side lineage
clause gates that path. A merge has no caller, so a story with no implementer dispatch or
no verifier dispatch is refused rather than falling through to the `{:ok, story}` that is
correct where a live caller is being judged and vacuous here.

## Fail closed, everywhere

A missing, empty, garbage or checksum-mismatched trigger document; an unknown repository; a
project with no intake source; an unreachable, rate-limited or truncating GitHub; a diff
that did not parse; a stale trigger at either ref; an unverified or custody-unattributed
story — every one refuses. There is no input and no failure that produces an `:allow` on
incomplete knowledge, and a refusal carries the **whole** list rather than the first item,
so fixing one reason does not mean running the gate again to discover the next.

**Gate B runs twice**, over the repository's file list at the head **and** at the merge
base, OR-ed. A pattern matching at one ref and not the other is drift — the rename that
silently unguards a trigger is precisely where the two disagree.

**The repository is never taken from the caller.** It is resolved from the story's project
through its GitHub intake source; no source, or more than one, refuses rather than
guessing. Gate B's trigger list is keyed by repository, so a caller that could name one
could be judged against another repository's triggers.

**An already-merged pull request is `already_merged`, not a refusal.** The outward effect
has happened, so a caller that crashed between merging and recording the merge adopts the
sha instead of being told its own completed merge is an escalation.

## The one write, and no second write path

A refusal takes `{:ci, :escalated, :merge_gate}` through `Loopctl.Delivery.Stages`, the only
writer of `story_stages`. An allow writes nothing: the caller performs the merge and then
advances `{:ci, :merged}` carrying the sha GitHub returned, because the merge commit does
not exist until the merge happens.

## SOUL rule 9, answered in the code

| question | answer, and where it lives |
|---|---|
| partition | GitHub unreachable, rate-limited or truncating is a `:refuse` naming the reason. There is no error path that merges. |
| where the verdict lives | Nowhere. It is recomputed from the forge and the database on every ask and never cached as authority; what persists is the consequence, the escalation. |
| retries | `evaluate/3` writes nothing, so the same pull request at the same head gives the same answer and a re-run after a new commit judges the NEW diff. `enforce/3`'s write is the stage machine's compare-and-set: a repeat finds the row already `escalated`. |
| slow connection | 2s connect, 5s receive, `retry: false` on every GitHub call, all made BEFORE any transaction opens, so a slow forge never holds a pooled connection. The caller sees a refusal naming the timeout. |
| where it runs | On whatever node serves the call. It owns no process and caches nothing, so there is no owner to find after a restart — every fact comes from Postgres or the forge. |

## Endpoint

`POST /api/v1/stories/:id/merge-precondition`, `exact_role: [:orchestrator, :user]` behind
`RequireHumanAnchor`. The same gate `review-complete` is behind and for the same reason: an
`:agent` key is the implementer's shape, and an implementer asking whether its own work may
merge is the self-approval the product exists to prevent. A refusal answers **200** — it is
an answer, not a request error — and has already escalated the story by the time it does.

## Testing

44 pure `judge/1` cases, 15 adapter cases against `Req.Test` bytes and 8 custody cases run
`async: true`. The end-to-end and controller suites are `async: false` with committed rows,
and their moduledocs say why: under the Ecto sandbox `Repo` and `AdminRepo` are separate
owners with separate transactions, so a row one wrote is invisible to the other and its
foreign keys fail — measured, a sandboxed `AdminRepo` tenant makes a `Repo` insert into
`story_stages` fail `story_stages_tenant_id_fkey`.

Pre-commit hook, verbatim: `9791 tests, 0 failures (88 excluded)`.

## Mutation testing

`~/workspace/claude-config/bin/mutate.sh <file> --old ... --new ... -- env -C <worktree>
MIX_TEST_PARTITION=_mergegate mix test <file>`. **23 mutations, every one exit 0** — the
check went red under the mutation and passed without it. Both halves the rule asks for:
the mechanisms, and the **wiring** that decides whether they run at all.

| # | what was broken | kind | exit |
|---|---|---|---|
| 1 | hard bound on files (`> 12` to `> 99_999`) | mechanism | 0 |
| 2 | hard bound on changed lines (`> 1000` to `> 999_999`) | mechanism | 0 |
| 3 | fail-closed fact handling (an `{:error, _}` fact became `{:ok, :mutated}`) | mechanism | 0 |
| 4 | Gate B's second run read the HEAD files instead of the merge base | **wiring** | 0 |
| 5 | the OR of the two Gate B runs (`worst/2` ignored the second) | mechanism | 0 |
| 6 | the custody call itself (`Progress.merge_custody_status(story)` became `:ok`) | **wiring** | 0 |
| 7 | `custody_reasons/1` returned `[]` for a refusal | mechanism | 0 |
| 8 | `missing_verifier_dispatch` clause disabled | mechanism | 0 |
| 9 | `not_verified` clause disabled | mechanism | 0 |
| 10 | Gate A's reasons dropped from the verdict | **wiring** | 0 |
| 11 | `:effect_proof_required` (a `:prove_effect` with no proof) returned `[]` | mechanism | 0 |
| 12 | the escalation edge became `{:ci, :implementing, :ci_red}` | **wiring** | 0 |
| 13 | the controller's `exact_role: [:orchestrator, :user]` became `role: :agent` | **wiring** | 0 |
| 14 | the adapter's file-list truncation guard (`> 100` to `> 100_000`) | mechanism | 0 |
| 15 | the adapter's status allowlist accepted `"unchanged"` | mechanism | 0 |
| 16 | the route path (so nothing reached the controller) | **wiring** | 0 |
| 18 | `{:pr_not_open, state}` returned `[]` | mechanism | 0 |
| 19 | Gate B's reasons dropped from the verdict | **wiring** | 0 |
| 20 | `missing_implementer_dispatch` clause disabled | mechanism | 0 |
| 21 | the reuse of `verify_recorded_separation/2` replaced with `{:ok, story}` | **wiring** | 0 |
| 22 | the escalation reason's 4000-character truncation | mechanism | 0 |
| 23 | repository resolution returned a hard-coded repo instead of refusing | mechanism | 0 |
| 24 | the adapter's truncated-git-tree refusal returned `{:ok, []}` | mechanism | 0 |
| 25 | the `RequireHumanAnchor` mount disabled | **wiring** | 0 |
| 26 | the egress chokepoint allowlist entry renamed | **wiring** | 0 |

(Numbering follows the order the mutations were written; 17 was dropped as unmeasurable —
a `config/config.exs` default the test environment overrides, so no test can see it move.)

## Judgement calls the brief did not settle

- **The endpoint's role gate is `exact_role: [:orchestrator, :user]`**, matching
  `review-complete`. An `:agent` key — the implementer's shape — is 403'd before any
  controller code runs.
- **The repository is server-resolved from the story's intake source**, not taken from the
  caller. The brief did not say where it comes from; a caller-supplied one could name a
  different configured repository and be judged against its triggers.
- **The behaviour returns a `DiffNames.parse/1`-SHAPED result**, fed to
  `DiffNames.merge_input/2` untouched, rather than raw `-z` bytes. `merge_input/2` is the
  single funnel, so a diff that did not parse becomes Gate B's unreadable-diff marker on
  both the GitHub path and any future path with a real checkout.
- **`already_merged` is a third decision**, not a refusal. Refusing a merge the forge had
  already performed would escalate a completed merge and strand the caller.
- **A `:prove_effect` outcome with no proof is a refusal**, and a proof of the wrong shape
  is a FAILED proof rather than an absent one.
- **Gate A's inputs are still caller-supplied** (`trio_outputs`), because triage does not
  persist them yet. Gate A escalates on anything that is not exactly three well-formed
  outputs, so a forgotten value fails closed — but a caller that fabricates three clean
  outputs clears Gate A. Reading them from a persisted triage record is the follow-up, and
  it belongs to the triage story that creates one; it blocks nothing here, because Gate B,
  the hard bound and the custody check are all server-derived.
- **`enforce/3` requires `:actor_lineage`**, because entering `escalated` is a chained
  transition and `Stages.advance/4` refuses one that does not declare it.
- **`Loopctl.Tenants.TierCapabilities` gained the new controller** under
  `chain_of_custody` — its drift guard caught the missing entry, which is the guard working.
- **The egress chokepoint allowlist and `docs/egress-guard.md` both gained a row** for the
  GitHub adapter (operator-plane, fixed vendor host, GET only, no tenant-supplied URL).

## Not done here

`GITHUB_TOKEN` now gates merges as well as verification — unset, an exhausted anonymous
quota escalates every delivery-loop change rather than merging anything. Its row in
`deploy/FLY_SECRETS.md` says so. **No new environment variable.**


mkreyman:
## Round 1: all 9 findings fixed — `0a76c31`

Hook line, verbatim: `9814 tests, 0 failures (88 excluded)` → `✅ All pre-commit checks passed!`
(pre-push e2e also green). Credo `--strict` and Dialyzer clean.

The two HIGHs changed the shape of the thing, so they are worth reading before the table.

### H1 — an ALLOW is now a recorded fact

`already_merged` discarded Gate A and custody, never reached the hard bound, Gate B or the
escalation, and answered "already merged, no reasons" for a merge performed around the gate
or after it refused. The last gate before an outward effect was ratifying an effect it never
licensed.

`enforce/3` now writes `story_stages.merge_gate_allowed_sha` — the head it judged — through
`Stages.record_effect`, the existing one writer of that table, as a side-effect identity like
the rest. `record_effect`'s idempotence is exactly the semantics an allow needs: the same
head twice is the same allow; a DIFFERENT head is `:effect_conflict` rather than a silent
re-grant. **A write that does not land turns the allow into a REFUSAL**, because an allow
nobody recorded is an allow nobody can later account for.

`already_merged` splits three ways:

| the forge merged… | verdict |
|---|---|
| a head a recorded allow names | `:already_merged` — adopt `merge_sha` |
| a head with NO allow, or an allow for another head | `:refuse`, `{:ungated_merge, sha, why}` — the sha is on the verdict AND in the escalation reason, so the fact is not lost |
| nothing (a nil `merge_commit_sha`) | `:refuse` — `advance/4` refuses a `merged` transition naming nothing, so `already_merged` would strand the caller (finding 7) |

The allow is cleared with `head_sha` by every edge that clears it
(`StageMachine.clears/3`), so a later, unjudged head can never inherit it. **Two existing
drift guards caught this change** — `stage_machine_test`'s `clears/3` assertion and
`stages_test`'s replay-every-effect loop — and both were updated rather than relaxed.

### H2 — a transient forge fault is not a verdict

Escalating takes `ci → escalated`, which is human-only, so one 5s timeout parked a story
until Mark acted. A transient fault is now a third outcome, `:unevaluated`: nothing decided,
nothing transitioned, the caller retries, and **HTTP 503** rather than 200 so it cannot be
mistaken for an answer.

`transient?/1` is the whole classification and is deliberately narrow — transport, a 5xx, a
429, and the 403 GitHub answers a rate limit with. A 404, a 401, a body that does not parse
and a truncated list are configuration or a contract change, and a human IS the right answer
to each, so they still escalate. **No retry was added inside the call**; the wall-clock
ceiling stays as documented, and the caller owns the retry.

### The rest

| # | fix |
|---|---|
| 3 | The judged head must equal the recorded `stage.head_sha` — the head CI ran on and the story was verified at. A mismatch, or a nil recorded value, is `:head_moved` and takes `{:ci, :implementing, :base_moved}`: new commits are ordinary work, not an escalation. |
| 4 | The escalation reason is bounded in CODEPOINTS (matching Postgres `char_length` and `Stages.bounded_text/2`), with a 100-codepoint margin. A grapheme bound let an NFD path fail the CHECK and leave the story at `ci` with nothing recorded. |
| 5 | A passing `effect_proof` can no longer produce an allow — `:prove_effect` escalates whatever the proof says, and the proof is still judged and recorded. Gate A's inputs cannot move server-side yet, so every verdict carries `gate_a_inputs: caller_asserted` and every escalation reason opens with it. Both moduledocs name what closes each properly: the design's server-side Gate B harness, and triage persisting its verdict. |
| 6 | `head_sha` goes through `ref/1` before it is spliced into the `/compare/...` path, like every other ref. |
| 8 | `inputs/1` and `gate_b/2` collect EVERY failure. `:not_attempted` is dropped — it is the consequence of a missing input listed alongside it, never a second fault. |
| 9 | `redirect: false`, as `ReqDelivery` pins it. A renamed or transferred repository would otherwise be read under its new name while the triggers stay keyed to the configured one. |

### Migration

`story_stages` gains `merge_gate_allowed_sha` (text, sha-shaped CHECK). No manual step.
Existing rows get NULL, so an already-merged pull request in flight at deploy escalates as
ungated rather than being adopted — deliberate, and it clears the next time the loop runs
the gate.

### Mutation testing — 18 new, every one exit 0

Same invocation as round 1. Both halves again: the mechanisms, and the WIRING.

| # | what was broken | kind |
|---|---|---|
| R1 | `{:ungated_merge, _, :no_recorded_allow}` → `[]` | mechanism |
| R2 | `{:ungated_merge, _, {:allow_for_other_head, _}}` → `[]` | mechanism |
| R3 | `:merged_without_sha` → `[]` | mechanism |
| R4 | the head-match comparison → `[]` | mechanism |
| R5 | `:head_sha_not_recorded` → `[]` | mechanism |
| R6 | `transient?({:github_unreachable, _})` → `false` | mechanism |
| R7 | `transient?(_)` → `true` (a 404 wrongly transient) | mechanism |
| R8 | `:unevaluated` deciding first → a refusal instead | **wiring** |
| R9 | a caller-asserted passing proof allowed again | mechanism |
| R10 | `input_reasons/1` truncated to one reason | mechanism |
| R11 | the `record_effect` call replaced by a fake success | **wiring** |
| R12 | a failed allow record no longer became a refusal | mechanism |
| R13 | `:head_moved` took the escalation edge instead of `:base_moved` | **wiring** |
| R14 | `:unevaluated` dropped from the "transitions nothing" clause | **wiring** |
| R15 | the codepoint reason bound → `40_000` | mechanism |
| R16 | `clears/3` no longer cleared the allow with the head | **wiring** |
| R17 | the adapter's `ref/1` on `head_sha` bypassed | mechanism |
| R18 | `redirect: false` → `true` | mechanism |

### Judgement calls this round did not have settled for it

- **The allow lives on the stage row, not an events row.** It is bound to a head exactly like
  the other identities, it goes through the one existing writer, and `record_effect`'s
  idempotence and `:effect_conflict` are the semantics an allow wants for free. An events row
  would have needed a new writer and its own conflict rule.
- **`:unevaluated` WINS over a custody or Gate A refusal** when both are true. The reasons
  still carry everything known, so nothing is hidden — but the decision says there is no
  verdict yet, because escalating on a blip is the failure H2 is about.
- **`:already_merged` transitions nothing** (the caller's next act is recording the merge it
  adopted), and so does `:unevaluated`.
- **A codepoint prefix can split a grapheme cluster.** Cosmetic, still valid UTF-8, and the
  trade is against an escalation that will not write at all.
- **`evaluate/3`'s `:allow` is a PREVIEW, not an authorisation** — only `enforce/3` records
  one. Said in the moduledoc and on the endpoint.
- **A `:prove_effect` outcome now has no path to an allow at all.** That is stricter than the
  design's step 4, deliberately, until the harness exists to make the proof loopctl's
  measurement rather than the caller's claim.

mkreyman:
## Round 2: all 5 findings fixed — `1699db0`

Hook line, verbatim: `9831 tests, 0 failures (88 excluded)` → `✅ All pre-commit checks passed!`
(pre-push e2e green). Credo `--strict` and Dialyzer clean.

### H1 — a bare 403 no longer means "retry", and nothing retries for ever

GitHub answers **both** a rate limit and a permanent permission denial with 403. A
fine-grained token with pull-request read but no contents read 403s both tree calls for ever,
so every evaluation answered `:unevaluated`, nothing transitioned, no escalation was ever
written and the loop retried indefinitely — the outcome the 404/401 classification exists to
prevent, reached through the one status assumed benign.

**Both remedies, as asked.**

*(a) Tell them apart.* `get/2` keeps the response now. `x-ratelimit-remaining: 0` or a
`retry-after` means a limit → `{:github_rate_limited, status, retry_after}`, transient.
Neither present → `{:github_api_error, 403}`, which escalates as configuration. A 429 is a
limit whatever its headers say. An HTTP-date `Retry-After` is deliberately **not** parsed into
a number — the caller's own floor beats a misparsed date.

*(b) The backstop, missing anyway.* Consecutive `:unevaluated` results at one head are
counted, and past `max_consecutive_unevaluated/0` the gate escalates with a reason naming
the fault.

Where the count lives, and why not `attempts`: that map's keys are EDGE names
(`StageMachine.counted?/1`), and a non-edge key would break what "attempts" means. It cannot
be a side-effect identity either — a counter is the one thing `record_effect/5` cannot hold,
since a second, DIFFERENT value there is `:effect_conflict` by design. So it is a new
`story_stages` column with its own writer in `Stages` (the one writer of that table), fenced
by the claim epoch, refused off `ci`, and leaving an event so a story going quiet is on the
record rather than merely absent.

It is keyed to the **recorded** head, not the forge's: when the pull request cannot be read
at all there is no forge head, and keying on that would collapse every story's count onto
`null`.

### The rest

| # | fix |
|---|---|
| 2 | The transient check is scoped to the facts the DECISION consumes — `merged? true` consumes only the pull request — and the merged path fetches no file lists at all. It was firing two identical `/git/trees` calls (the adapter reports a merged pull request's merge base as its head) whose results were discarded, and a rate-limited one could suppress the ungated-merge escalation, the loudest signal this module produces. |
| 3 | A verdict converted to `:refuse` re-enters `act/4`, so it takes the same escalation path as any other refusal. Previously `act/4` had already dispatched on `:allow`, so the loop got `refuse` over HTTP 200 while the story sat at `ci` with nothing recorded — and the `:effect_conflict` variant repeated that answer for ever, since the epoch does not move across a `ci → implementing → ci` cycle. The recursion terminates: the second call carries `:refuse`. The integration test asserts the row's stage now, and there is a second test for the stale-epoch case where BOTH writes are fenced and both failures are reported rather than an escalation being claimed that was not written. |
| 4 | Every 503 carries `Retry-After` — the forge's own delay when it named one, else a stated floor of 30s. Both are in the operation description, so the contract never reads as "retry at will". |
| 5 | The `GITHUB_TOKEN` row now says what the code does: an exhausted anonymous quota answers `unevaluated`/503 and the loop RETRIES, escalating only once the same story has been unevaluable at one head several times running. Either way nothing merges. |

### Migrations

`story_stages` gains `merge_gate_unevaluated` (jsonb, `{}` default, object CHECK), and the
`story_stage_events` event-name allow-list gains `merge_gate_unevaluated`. No manual step.

### Mutation testing — 11 new, every one exit 0

| # | what was broken | kind |
|---|---|---|
| S1 | `403` put back in `@transient_statuses` | mechanism |
| S2 | the adapter's headerless-403 clause made unreachable | mechanism |
| S3 | `x-ratelimit-remaining: 0` no longer read as exhausted | mechanism |
| S4 | the `retry-after` delay dropped | mechanism |
| S5 | the merged path consuming all three forge facts again | **wiring** |
| S6 | the backstop call replaced by a pass-through | **wiring** |
| S7 | the backstop bound raised to 100 000 | mechanism |
| S8 | the per-head reset removed (count ignored the head) | mechanism |
| S9 | `note_unevaluated`'s `ci`-only guard disabled | mechanism |
| S10 | a converted refusal returning instead of re-entering `act/4` | **wiring** |
| S11 | the `Retry-After` header dropped | mechanism |

Running total across three rounds: **52 mutations, all exit 0.**

### Judgement calls this round did not have settled for it

- **An HTTP-date `Retry-After` yields `nil`, not a guess.** The endpoint's floor applies. A
  misparsed date is worse than no number.
- **`max_consecutive_unevaluated` is 5.** Small on purpose: escalating a story that would
  have recovered costs one human glance; not escalating costs a story nobody hears about
  again.
- **A count that cannot be written leaves the verdict `:unevaluated`** and reports
  `{:unevaluated_not_counted, reason}`. The backstop is a safety net, not a second way to
  refuse.
- **The per-head reset is tested at `Stages.note_unevaluated/4`, not through the endpoint.**
  Moving the recorded head end-to-end needs a full `ci → implementing → … → ci` cycle; the
  reset is that function's own property and is mutation-tested there (S8).
- **`recorded_head_sha` is now on the verdict and in the response** — a caller can see what
  CI ran on without a second call, and it is what the count keys on.

mkreyman:
## Round 3 (ceiling): all 6 findings fixed — `bafcbcb`

Hook line, verbatim: `9843 tests, 0 failures (88 excluded)` → `✅ All pre-commit checks passed!`
(pre-push e2e green). Credo `--strict` and Dialyzer clean.

### 1 + 2 — one mechanism, and round 2's two fixes combined into it

`x-ratelimit-reset` was discarded, so a **primary** rate limit (the anonymous-quota case
`FLY_SECRETS` documents) produced no delay, the endpoint's 30s floor applied, and the new
5-count bound escalated the story in ~3 minutes against a window of up to an hour. The delay
now comes from `x-ratelimit-reset` when `Retry-After` does not parse, capped at the forge's
longest window plus slack, and ignored when it points into the past.

**Presence decides the class; only the parse decides the delay.** A 403 whose only
rate-limit signal is an HTTP-date `Retry-After` — legal, and the shape that says "come back
later" most plainly — was falling through to permanent-permission-denial and escalating at
once. `Retry-After` present, or `x-ratelimit-remaining: 0`, classifies. `x-ratelimit-reset`
never classifies, because it rides on ordinary responses too; it only supplies the number.

The round-2 HTTP-date test also set `x-ratelimit-remaining: 0`, so it exited through the
exhaustion branch and never covered this. Fixed, plus negative, unparseable, reset-only,
reset-in-the-past, and reset-without-a-limit-signal cases.

### 3 — the count that never cleared

`merge_gate_unevaluated` was cleared by nothing: not a successful evaluation, not the
escalation it triggered, not any edge. After the backstop fired, a `human_resolution` back to
`queued` left it standing and one blip at the same commit escalated again.

Cleared now on an evaluation that produced a verdict (`Stages.clear_unevaluated/3`, a no-op
when there is nothing to clear) **and** by every edge that clears `head_sha`. The three
head-bound fields are one list — `StageMachine.head_keyed/0` — and the drift guard walks the
whole transition table asserting that any transition clearing `head_sha` clears all of them.
They were added one at a time, to one clause at a time, which is exactly how the next one
gets forgotten. The column is nullable now, because clearing a field here means `NULL`.

### 4, 5, 6

| # | fix |
|---|---|
| 4 | The `head_moved` branch is scoped like the merged one, at BOTH the fetch and the judgement: it reads neither file list, so a rate-limited tree call could mask it and — past the bound — escalate a story for an ordinary push instead of returning it to `implementing`. |
| 5 | The down migration deletes the `merge_gate_unevaluated` event rows before restoring the old allow-list. Verified against a COMMITTED event row: rollback and re-migrate both clean. |
| 6 | **Three** fields were emitted and undeclared, not two — `gate_a_inputs` as well. The round-2 edits meant to declare them silently matched nothing. All three are declared, and there is now a guard that fails when the rendered payload and the published schema disagree **in either direction**, because this drifted twice with nothing noticing. |

### Mutation testing — 7 new, all exit 0, and one that had to be earned

| # | what was broken | kind |
|---|---|---|
| T1 | the reset-on-success call replaced by a pass-through | **wiring** |
| T2 | `clear_unevaluated` no longer nilled the column | mechanism |
| T3 | `merge_gate_unevaluated` dropped from `@head_keyed` | **wiring** |
| T4 | `Retry-After` presence no longer classified | mechanism |
| T5 | the reset-derived delay dropped | mechanism |
| T6 | the `head_moved` fact scoping removed | **wiring** |
| T7 | the `head_moved` fetch skip removed | **wiring** |

**T6 came back exit 1 first** — the check could not fail. The fetch skip (T7) makes that
clause unreachable through `evaluate/3`, so nothing could observe it. Rather than delete a
guard that is real defence in depth, or keep one that cannot go red, it is now falsifiable
against `judge/1`, whose contract holds for any facts rather than only the ones `gather/3`
builds today. Re-run: exit 0.

Running total across four rounds: **59 mutations, all exit 0.**

### Judgement calls this round did not have settled for it

- **`x-ratelimit-reset` supplies the delay but never the class.** It appears on ordinary
  responses, so using its presence to classify would call a permission denial a rate limit —
  the exact inversion round 2 fixed.
- **Reset delays are capped at 3900s** (the hourly window plus slack) and a reset in the past
  yields `nil`. Neither tells a caller anything it should sleep on.
- **A failed clear does not withdraw the verdict.** It is tidying that did not land: the worst
  it costs is an early escalation on a later fault, and refusing a merge the gate allowed
  would cost more. Reported as `{:unevaluated_not_cleared, reason}`.
- **`head_keyed/0` is separate from `effects/0`** because `merge_gate_unevaluated` is a
  counter, which `record_effect/5` structurally cannot hold.
- **The unevaluated bound still counts EVALUATIONS, not elapsed time.** A caller that ignores
  `Retry-After` can burn five in seconds — and escalating a caller that ignores the contract
  is the right outcome, so this is deliberate rather than unfinished.

PR #836, opened 2026-09-15
Publish every bound the limits table claims to hold, and stop claiming a BMP-only fit

mkreyman:
Two defects in the 1.7.0 that already merged, both found by the `loopctl-runner` maintaining session building an emitter against it. The second was already written down in the knowledge base before I wrote the guard that missed it.

## The limits table was incomplete and read as complete

`field_bounds/1` handled a flat string and an array **of strings** and nothing else, so `limits.triage_verdict.fields` omitted `contradicts` and the **entire draft story sub-object** — every one of which is bounded in `$defs`.

`RunnerStory` is flat, so `limits.story.fields` was complete. That is what made this dangerous rather than merely missing: it looked like a **precedent**. A runner author reading bounds off the table the way that precedent prescribes got no bounds at all for the story it drafts, wrote a 1,400-character description into a 900-character field, and had the verdict refused for a rule the table did not contain.

Same defect as a `maxLength` that is not what binds: something that reads as the bounds and is not all of them.

It now walks every bounded field at every depth — nested objects and arrays of objects included — and keeps `max_item_length` for the string case so a runner reading the old table still finds what it read before.

## And there were three copies of it

`RunnerStory`, `RunnerTriage` and `RunnerTriageVerdict` each carried an identical `limits/0` and `field_bounds/1` — precisely the drift this contract's comments spend paragraphs warning about, created by me two commits earlier. They were not identical for long: the fix above had to be applied three times, and `bin/mutate.sh` refused the ambiguity, which is how I noticed. One implementation now, `RunnerContract.Limits`, taking the schema and the cap.

## The every-field-at-maximum guarantee was BMP-only

Round 3 of #835 established that a payload inside every published bound is accepted, and measured it with ASCII. `ByteRule` charges 12 bytes for a character outside the BMP against 6 inside, so the same verdict is **~89,584 bytes against a 48,000 cap** once the text is astral — and an emoji in a drafted story title is ordinary, not adversarial. The guarantee was true only for text nobody promised to send.

**The KB already said this**, from contract 1.1 (`b05ce162`): *worst-case frame tests must use control and astral characters at every max length, not plain ASCII.* I wrote the guard without searching for it — a rule-0.5 failure — and a peer session rediscovered it by measurement.

**The caps are deliberately NOT resized.** Sizing every field for the worst case of every field at once would halve a draft story's description to defend a case no session produces, and the byte rule is already a worst-case-encoder bound, so that is conservative twice over. What changes is the claim:

- the two fits-tests are renamed to say **BMP**,
- a third test pins the astral cost as a measured fact in the suite,
- both schema descriptions now say plainly that **the object cap is what binds**, that the rule charges double outside the BMP, and that a session must check the byte rule against `max_bytes` rather than the field lengths.

That is what the runner's emitter will do.

## Falsifiable

Two mutations, each exit 0:

    the nested-object branch of field_bounds matches a type that cannot occur
      1 red: the draft story's bounds vanish from the table again.

    the astral test's fill reverts to ASCII
      1 red: it stops measuring the thing it exists to measure.

Worth recording: the **first** version of that astral test filled only the top-level strings and reported 45,880, which looked like it nearly fitted. The fill now reaches every string at every depth, and the real figure is 89,584.

Full gate green at 10,591.

## Separately verified, no change needed

The runner found that *its* validator quoted offending values into error text, so a malformed `untrusted` field could have put the reporter's words into its journal and back across the wire. I checked the mirror here by measurement — six probes across `cast_dispatch/1` and `cast_triage_verdict/1`, over-length, wrong-type and nested — and every one came back clean: OpenApiSpex names the constraint and the type, never the value. No log line or reply interpolates a payload or an `untrusted_*` field either.

Review gate still running.


mkreyman:
Superseded by #838, which carries the identical change cut from master instead of from #837's pre-review commit.

Closed for a specific hazard rather than tidiness. This branch was stacked on `89c7dbc`, the commit #837 was at BEFORE its review. #837 has since taken nine review fixes including two HIGHs — one where `target_epic_id` was unreachable through every API path, one where a failed stage-open reported success. After #837 squash-merges, git's lowest common ancestor for this branch is still `89c7dbc`, so merging it could restore the pre-review versions of those files with a clean merge and no conflicts, silently reverting every one of those fixes (KB `33edd3e3`).

#838 is cut from master and touches a disjoint file set, so it has no relationship to #837 and the two can merge in any order.

PR #854, opened 2026-09-15
Contract 1.10.0: triage is dispatchable, and loopctl sends it (#803)

mkreyman:
**The hop the loop did without.** Intake promoted a reported issue to a story at `detected`, the verdict path could apply a triage verdict — and nothing asked a runner to do the triage in between: `dispatchable_kinds` was `implement` alone and nothing in `lib/` built the dispatch. A real ticket stopped at `detected` for ever, and the trio ran only where a person ran it.

## The interlock moved because both its conditions are met

The contract documented exactly when `triage` could become dispatchable, and that reasoning is kept rather than deleted, because it is what says when it may move again. What held it back was the other end: no runner accepted the kind, so sending it would spend a dispatch on a refusal and, against an UNDECLARING runner, write a permanent `kind_not_supported` for that machine.

Both halves are answered. The runner composes its three triage lenses and emits a `triage_verdict` (1.9.x). And an undeclaring runner is still protected, because **`implied_by_silence` stays `implement` ALONE** — a machine built before the `kinds` field existed is sent exactly what it was sent before, and only a runner that DECLARES `triage` on join receives one. That asymmetry is the whole safety of the bump and must not be tidied away; there is a test that says so.

## What sends it

`Loopctl.Delivery.TriageDispatcher` + `Loopctl.Workers.TriageDispatchWorker` (every minute, off with the driver under the same `:dispatch_driver_enabled`). It selects stories at `detected` that came from an intake record — the payload IS the reporter's words, so a story without one has nothing to triage — fairly across tenants, bounded per pass.

**It claims nothing.** A triage session decides whether the story is work at all; claiming would mark it as being worked by the machine that is only deciding that, and would put it at `claimed`, which is the stage an implement session reports from. The dispatch carries the story's current `claim_epoch`, which is what the ledger fences and what the verdict's transitions are checked against when it comes back.

**The `dispatch_id` is derived** from the story and its epoch rather than generated: a story stays at `detected` until its verdict arrives, so it is a candidate of every pass in between, and a generated id would start a fresh triage session every minute on the same ticket. A busy answer — a contended reservation, a full runner, the tenant's admission cap — is `:deferred` at info, apart from `:blocked`, which means somebody has to change something.

**Its own budgets**, `TRIAGE_WALL_CLOCK_SECONDS` and `TRIAGE_MAX_TURNS`, with no default: a triage run reads a ticket and answers, so inheriting the implement numbers would hand a reading session a day of wall clock.

## Mutation testing

Four runs, each exit 0 (red under the mutation): equalising `implied_by_silence` with `dispatchable`; dropping the intake-record predicate; widening the stage predicate to include `queued`; generating the dispatch id instead of deriving it.

Also in this branch, from the merge with master: the `place_dispatch` MCP tool accepts `kind: "triage"` again (mcp-server 2.96.0), and the contract digest guard from #848 is re-pinned for 1.10.0 — which is the guard working, not a workaround.

The review gate is still to run on this.


mkreyman:
Round 1's eight findings are all fixed in-branch (cedd5829) — the CRITICAL budget-ceiling crash, the three HIGHs (all one cause: the derived dispatch_id, now generated fresh with the exclusion moved into the candidate query as a predicate on the unreleased ledger row), the triage_too_large escalation, the place_dispatch kind enum, the detected index, and the transient/blocked split. Five mutations at the wiring, all exit 0. The commit message carries the per-finding reasoning.

Issue #849, opened 2026-09-15
stale_stage refuses with no row state, so the runner brute-forces three froms — and no test on either side covers a real dispatch's first stage report


mkreyman:
Two defects found while diagnosing why every runner stage report on minis was refused `stale_stage`. The refusals themselves were CORRECT — the story had never been claimed, because that dispatch went out by production RPC rather than through `Placement.place/4`, so the row sat at `queued` while the runner reported `claimed -> worktree`. That part needs no code change; it needs the placement path turned on.

These two are what made a five-minute diagnosis take an hour, and they will do it again.

## 1. The `stale_stage` refusal tells the runner to re-read and gives it nothing to read

`lib/loopctl/api_spec/runner_contract.ex:171` prescribes the remedy:

> ELSE is `stale_stage`: re-read the story and send the transition that applies.

But `stale_stage` is in the `@verbatim` list at `lib/loopctl_web/runner_channel/refusal.ex:97-101`, whose clause is `%{reason: Atom.to_string(reason)}` — so the reply carries the reason and nothing else. No `stage`, no `claim_epoch`, no `lock_version`, no `attempts`, though `RunnerStages` has the row in hand when it refuses.

The runner cannot follow the contract's instruction, so it does the only thing left: it brute-forces `@fallback_froms = ["implementing", "worktree", "claimed"]` (`loopctl-runner/lib/loopctl_runner/control.ex:1760`). On 2026-09-14 at 14:20:32 all three were refused in turn — three round trips that still could not name the row's actual stage, and an operator reading the journal could not either.

The `RunnerStages` moduledoc already says what the reply is FOR:

> The reply carries the row's `stage`, `claim_epoch`, `lock_version` and `attempts`, which is how a session that lost its acknowledgement learns where the story actually is without a second endpoint.

That is true of the `{:ok, row}` path and false of this refusal. **Carry the row on `stale_stage`** — it is the one refusal whose entire remedy depends on knowing where the row is — and the runner's fallback list can be deleted rather than extended.

## 2. Neither side has a test that a dispatched story can report its first stage

The mismatch survived because both suites assume it away:

- loopctl, `test/loopctl_web/runner_channel_stage_test.exs:80-85`: calls `Runners.dispatch/3` and then **manufactures the stage row by fixture at `:implementing`** — handing the runner a state production never created.
[...]

PR #855, opened 2026-09-15
Contract 1.11.0: a stale_stage refusal carries the row (#849)

mkreyman:
Closes the first half of #849.

## The defect

The contract has prescribed the same remedy for `stale_stage` since 1.4.0:

> ELSE is `stale_stage`: re-read the story and send the transition that applies.

The refusal carried the reason and nothing else. `story_stages` has no runner-facing endpoint, by design, because the reply IS the read — so the instruction was unfollowable, and the deployed runner did the only thing left: it brute-forced `@fallback_froms = [implementing, worktree, claimed]`. On 2026-09-14 at 14:20:32 all three were refused in turn. Three round trips that still could not name the row's stage, and an operator reading the journal could not name it either.

`RunnerStages` had the row in hand at the moment of the refusal the whole time. `RunnerStages`' own moduledoc already claimed this was true.

## The change

- `RunnerStages.resolve_replay/3` returns `{:error, {:stale_stage, row}}`.
- `RunnerStages.row_state/1` is the ONE renderer for a stage row on the wire — `stage`, `claim_epoch`, `lock_version`, `attempts`, `effects`. The channel's `stage_ack/1` now calls it, and so does the refusal, so the ack and the refusal cannot name different fields.
- `Refusal.for_message({:stale_stage, row})` renders it beside `reason`.
- The BARE `:stale_stage` clause stays. `triage_verdict` is refused with the same code from `TriageVerdict`, where it means the story left `detected` and no transition applies at all — PERMANENT for that message, per `x-connection.permanent_errors`. Handing back a row there would suggest a retry the contract refuses.
- Contract 1.10.0, digest re-pinned, export regenerated. Additive on the wire: nothing a 1.9.x runner does breaks, and a runner holding a `from` fallback list can delete it.

## Verification

Three mutations, all at the WIRING rather than the mechanism, all exit 0 (went red):

    bin/mutate.sh lib/loopctl/delivery/runner_stages.ex --old '{:error, {:stale_stage, row}}' --new '{:error, :stale_stage}' -- mix test test/loopctl/delivery/runner_stages_test.exs test/loopctl_web/runner_channel_stage_test.exs
    bin/mutate.sh lib/loopctl_web/runner_channel/refusal.ex --old 'do: Map.put(RunnerStages.row_state(row), :reason, "stale_stage")' --new 'do: %{reason: "stale_stage"}' -- mix test test/loopctl_web/runner_channel_stage_test.exs
    bin/mutate.sh lib/loopctl_web/runner_channel.ex --old 'defp stage_ack(row), do: RunnerStages.row_state(row)' --new '<the old inline map, effects blanked>' -- mix test test/loopctl_web/runner_channel_stage_test.exs

New coverage: the refusal at the unit (`refusal_test.exs`, asserted against `row_state/1` rather than a restated field list), at the context (`runner_stages_test.exs`, asserting the row names `reviewing`), and ON THE WIRE (`runner_channel_stage_test.exs`, every field compared against the ok ack's).

## Second half of #849

The integration test that places through `Placement.place/4` and drives a real first stage report landed in #853. `loopctl-runner`'s `@fallback_froms` removal is that repo's half.


mkreyman:
Round 1's five findings are all fixed in-branch (9471c949). The two MEDIUMs were the substantive ones: x-connection.error_fields now publishes what every refusal carries beside its reason, per event and complete, and permanent_error_conditions gains the stale_stage stopping rule for a terminal stage. The three LOWs are fixed as described. Both mutations at the wiring, exit 0. Details in the commit message.

PR #856, opened 2026-09-15
The delivery loop can FINISH: verified -> done has a writer (#803)

mkreyman:
Closes the last unwritten edge of the delivery loop's main line, found by a systematic sweep of every transition `StageMachine` declares against what in `lib/` can actually write it.

## The defect

**Nothing in `lib/` wrote `verified -> done`.** A story that passed every gate — triaged, queued, claimed, implemented, reviewed, merged, deployed and verified — stopped one stage from the end of the line and stayed there for ever. `verified` had no writable edge out at all:

- not a runner source stage (`runner_source_stages/0` stops at `merged`),
- not in the `@session_escalated` set (`@in_flight ++ [:merged, :deployed]`),
- and `:human_resolution` LEAVES `escalated` rather than reaching it.

So not even an operator could move it.

This is the same defect as `triaged -> queued` (#847) and the same shape the comment above `@session_escalated` already records for `merged` and `deployed`. **It moved one stage forward each time a writer was added.** The comment saying "`deployed -> verified` stays for the control writer that does not exist yet" outlived that writer — `PostDeployVerification` writes it now — so the trap read as already-solved while it sat one stage further on. That comment is corrected, and `verified` is the end of the line.

## What `done` means, and why it is not `verified`

Two different statements, and the loop needs both:

- **`verified`** — the deploy was checked and it shipped. `resolution_verdict/1` maps `{:deployed, :verified, :forward}` to `:shipped`, and `Stages.advance/4` writes the `intake_issue_closures` outbox row in that same transaction. At `verified` **the reporter has been promised something and nothing has been said to them yet.**
- **`done`** — every outward obligation is discharged. That is already what `live_row/2` means by excluding `[:done, :failed]` from every claim and release.

The gap between them is a real interval: the drainer retries a transient forge fault with backoff, and during it "is this finished?" has the answer no.

## The settlement rule, and the class it must not strand

A verified story is `done` when its closure row is terminal — `:closed`, or `:abandoned` (the drainer gave up and left the row for a person; treating that as unsettled would hold the story for ever over an act that is never going to happen) — **or when there is no row at all**, because the story came from no intake record and owed nobody anything.

That second clause is the whole reason this is not simply "the drainer marks it done". Keying completion on the closure row alone strands every backfill and every API-created story at `verified` for ever, while the sweep reports a clean run — this very defect, one layer down.

## Not behind `:dispatch_driver_enabled`

Unlike the three delivery crons beside it, and deliberately. That flag gates STARTING sessions on somebody's machines unattended, which is what an operator is agreeing to. This worker starts nothing and spends nothing; it records that work already finished. Gating it would strand every in-flight story at the final stage the moment a fleet turned the driver off — the absorbing state, on a switch.

## Recorded rather than fixed

`{:verified, :failed, :budget_exceeded}` is one of **twelve `:budget_exceeded` edges the machine declares from every live stage that nothing in `lib/` writes.** They are reserved, not broken: budget exhaustion is already handled — the runner stops at `wall_clock_seconds`, `Capacity.heal/3` releases the reservation, and the claim reclaimer requeues over `:runner_lost` — and routing it to `:failed` instead would terminate stories that should be retried, which is a policy nobody has chosen. `completion_test.exs` is where that is now written down, because a reader counting declared edges out of `verified` finds two and wrongly concludes it was never absorbing.

## Verification

Five mutations, **three at the WIRING** rather than the mechanism, all exit 0 (went red):

    bin/mutate.sh lib/loopctl/delivery/completion.ex --old 'where: is_nil(u.story_id),' --new 'where: true,'
    bin/mutate.sh lib/loopctl/oban_config.ex --old '{"* * * * *", Loopctl.Workers.StoryCompletionWorker},' --new ''
    bin/mutate.sh lib/loopctl/delivery/completion.ex --old 'where: c.status == :pending,' (the anti-join keyed on existence instead)
    bin/mutate.sh lib/loopctl/delivery/completion.ex --old 'if pending_closure?(tenant_id, story_id) do' --new 'if false do'
    bin/mutate.sh lib/loopctl/workers/story_completion_worker.ex --old 'if errored == length(outcomes),' --new 'if false,'

No new operator variable, no migration, no manual step. The review gate is still running.


mkreyman:
Round 1's six findings are all fixed in-branch (f5d8fbb0). The one that mattered inverts the settlement rule: **`:abandoned` does not settle**, because `requeue_abandoned/1` reopens one and `done` is terminal — a misconfigured `GITHUB_TOKEN` would have had a whole batch of stories recorded as discharged a minute before the operator fixed it. Also corrected which race the re-check closes (a fresh row cannot appear at `verified`; a requeue can, and the re-check narrows rather than closes it — accepted deliberately, with the reason in the code). Plus the missing partial index, the missing `unique:`, benign races demoted to info and counted apart, the `@spec`'s missing field, and the phantom `:skipped` outcome.

Four mutations, all exit 0 — one first came back GREEN and needed a new test: swapping the re-check's predicate could not fail, because the only row that tells `!= :closed` from `== :pending` apart is an abandoned one, and the candidate query never hands one to the writer.

PR #859, opened 2026-09-15
An escalation reason is escaped before it is stored, and two false claims are corrected (#804)

mkreyman:
Three of the #804 audit's remaining findings. They are one subject: text a SESSION wrote reaching a permanent record with its hidden characters intact, and two places the code promised protection it did not have.

## The escape

`escalation_reason` is written by a model that had just read reporter text, and it lands in two places nobody can edit afterwards — the `story_stages` column an operator reads, and, on a chained transition, the tenant's append-only hash chain. A bidirectional override or a zero-width run in it is invisible in both, so the text an operator sees is not the text that was written, for ever.

`Stages.advance/4` now runs it through `Untrusted.sanitise/1`.

**Before the bound, deliberately.** Sanitising EXPANDS text — one bidi mark becomes eight characters — so a reason inside the cap can sit outside it once escaped. Bounded after, the column would take a value the check had never seen; bounded before, the caller is refused for a length it can measure. It also rewrites `opts` rather than a local, because `transition/6` reads `:reason` again from there, and a second read of the raw value is exactly how the escaped form would reach the event while the raw one reached the column.

## This reverses a documented decision, and the argument belongs in the open

`Escalations` said the reason is stored VERBATIM so an operator reads what was actually said. That intent is kept: `sanitise/1` does not change what was SAID — prose stays prose, which is why `escalation_block/1` still has work to do at the hop where the text meets a model — it makes hidden characters visible, which serves that same operator better.

**What decides it on its own is the NUL.** Postgres will not store one, so a reason containing a NUL was refused outright and THE ESCALATION WAS LOST: a session asking for a human got nothing, over one byte, and the story stayed where it was. It is now recorded visibly and the escalation lands.

The measured cost, in the test rather than in a claim: an emoji joined by zero-width joiners shows those joiners, the family sequence goes from 5 codepoints to 19, and a caller near the 4,000 cap therefore has less room than its raw length suggests. The boundary test measures against the SANITISED text, which is what the column holds.

## Two false claims, corrected in place rather than quietly deleted

- **`Untrusted`** said "the fence brackets never survive inside the data". They do not — but a LOOK-ALIKE does. The escape is a codepoint denylist rather than a confusable class, so `U+301A`/`U+301B` pass through unescaped and are near-identical to the reader this fence is written for. The block still holds (the line arrives `| `-prefixed and carries a nonce the writer could not know), so what was false is that layer's own promise, and it now names which two layers carry it.
- **`Stages`** said the JSON an operator reads "gets the raw value". It gets the stored value, which is now escaped.

## Not changed, and why

`escalation_block/1` still has no non-test caller. Its own moduledoc already says so in a block quote — *"this is a CONVENTION with no binding guard, and it is on the next story to make it one"* — and names the shape the binding takes. Nothing in `lib/` renders the reason into a prompt today because no such prompt exists, so there is nothing to bind it to. An earlier inventory of mine called this "a convention that reads as a mechanism"; that was wrong, and it reads as exactly what it is.

## Verification

Two mutations, both at the WIRING, both exit 0 (went red):

    bin/mutate.sh lib/loopctl/delivery/stages.ex --old 'opts = sanitise_reason(opts)' --new ''
    bin/mutate.sh lib/loopctl/delivery/stages.ex --old '(the call, placed before the reason is read)' --new '(moved after)'

The second is the ordering the expansion hazard turns on, so it is the one worth having.

Full suite green at 10,743 tests.

## Still open on #804 after this

Acceptance 2 — *"the implementer's prompt provably contains no verbatim reporter text"* — remains unproved. The existing boundary test proves the BUILDER never reads intake, which is weaker. The cheap version, from #857's review: seed the intake record, drive a verdict whose drafted fields are a verbatim copy of `untrusted_title`/`untrusted_body`, and assert no 6-token shingle of the body appears in `ImplementerInput.story_object/2`'s output. That is its own change and is next.


mkreyman:
Round 1's nine findings are all fixed (4834799c), and one of them was a REGRESSION I introduced rather than a pre-existing gap.

I moved the length bound onto the escaped text while the controller and the contract still bounded the raw text, so a 4,000-codepoint reason carrying one zero-width space passed the HTTP check and was then refused 422 — the escalation lost, which is the exact failure this PR exists to end, reintroduced one layer up. A caller cannot predict the escaped length without implementing loopctl's escape table, and the contract publishes that bound for runners to validate against, so bounding the escaped form publishes a number no client can honour. The order is inverted and the column's CHECK is widened instead (migration 20260921170000).

Also: the presence check ran on escaped text, so a reason of `<U+000C>` satisfied 'a reason is required' — now on the raw text. The escape is not injective and the moduledoc now says so. Two of my own tests were vacuous (the boundary pair did not straddle the bound; the NUL test checked only the column, not the event or the chain) and both now assert the discriminating case. Three stale claims corrected: the OpenAPI spec, the render comment, and the 422 message.

Two mutations at the wiring, both exit 0. Full suite green.

PR #860, opened 2026-09-15
Standing rule: an engineer agent writes the code, not the session's own Edit calls

mkreyman:
Mark's standing instruction, 2026-09-15, verbatim:

> make it a standing rule that you dispatch an engineer agent on high effort for any implementation and fixes. I see it over and over again that sessions doing direct editing are doing sloppy job...

This is the loopctl half — it binds delivery-loop **step 2** in this repo, and `CLAUDE.md` is the tier a dispatched subagent is guaranteed to read. The fleet-wide half is routed to claude-config as **issue #597**, because `~/.claude` cannot be written from a loopctl session and should not be.

## The rule

Dispatch `elixir-engineer` (or `typescript-engineer` under `mcp-server/`) on the highest-effort model for any change to CODE. The session keeps the loop — branch, reconcile the agent's output against the diff, commit, PR, review gate, merge — and the agent leaves its changes uncommitted, which is its correct end state under the Subagent Subordination Rule.

**The floor is named** so this does not become ceremony: version bumps, CHANGELOG entries, formatting, compiler-verified renames, regenerating checked-in artifacts, resolving a conflict already understood, and prose in a doc like this one. Anything that decides BEHAVIOUR is the agent's.

## The evidence is in the section, not left as a claim

One session, 2026-09-15, five PRs on the delivery loop, every change hand-edited. What `/code-review high` found after each was committed and pushed:

- three tests that could not fail, each written to pin a NEW guard, each caught only by `bin/mutate.sh` afterwards;
- a false claim deleted from one file and left standing in another **by the same commit**;
- a comment about the audit chain that was wrong, then rewritten wrong a SECOND time while being corrected;
- a change that made a case worse than before it — a bound moved onto escaped text while the caller-facing bound stayed raw, losing a conforming caller's escalation.

None is a typing error. They are judgement errors made at depth in a long session.

**A sixth landed while this very rule was being written**, and it belongs on the record: a `git add -u` swept `CLAUDE.md` into an unrelated security PR — the same sweep failure the deny-list on `git add -A` exists to prevent, which `-u` is not covered by. That is what this branch exists to undo, and it is why the commit message says so rather than quietly re-landing the file.

## What the dispatch actually changes

Stated precisely, because a rule adopted for the wrong reason is dropped for the wrong reason. The agent is the same model and can make the same judgement errors. What differs: a FRESH CONTEXT on one task rather than a session carrying hundreds of thousands of tokens of unrelated history — every failure above happened deep into a long session; a system prompt carrying this stack's conventions; and a read inserted between writing and committing, where today the same principal does both.

It does not replace the review gate.

**What would overturn it:** agent-written changes showing the same defect rate under review. Measure that rather than asserting it either way.

## Note on this PR itself

Written directly rather than dispatched, under the floor the section names — it is prose about process, with no behaviour in it.


mkreyman:
Round 1's ten findings are all fixed. The best of them is about this rule itself: **the floor was excluding half its own evidence.** It tested on 'does it change behaviour', which permits hand-editing prose — and two of the four defects the section cites are exactly that, a contract sentence and a code comment. A session following the rule as written would still have shipped both. The test is now 'does the change assert anything about how the code works'.

Also: the section was second-person and `CLAUDE.md` is the tier a subagent is guaranteed to read, so `elixir-engineer` would have received an instruction it has no Agent tool to obey; the remedy did not address the strongest evidence item (a fresh context writes an unfalsifiable assertion just as easily, so the dispatch prompt now has to carry `mutate.sh` with its exit); it contradicted step 1's worktree fan-out; two floor clauses were self-graded or unsafe; and I had the mechanics wrong — both agent definitions already pin `model: opus` / `effort: high` and an override can only downgrade, which I verified by reading the frontmatter rather than restating it.

Evidence now carries the PR and commit ids, and the overturn test has a baseline (~6 confirmed findings per PR across that batch) so it is runnable. Prose gets one round; this is that round.

PR #865, opened 2026-09-16
A placement that cannot give back the claim it made now escalates instead of logging (846.1)

mkreyman:
## The story it implements was stale, and that is the first thing to check

846.1 says *"a REFUSED dispatch must release the claim — today it parks the story at `claimed` for HOURS"*. **That was already true.** `Placement.undo_claim/5` releases inline on the refusal reply, and `placement_test.exs`'s "a refused push releases the claim it made" proves it end to end.

What actually happened on 2026-09-15 is narrower and worse: the release **ran and failed**, and nothing covered a failed compensation. The story sat at `claimed` for four hours behind a `Logger.error` on a box nobody was tailing — on a loop whose entire point is running unattended.

So this implements the incident rather than the story text.

## Which failures escalate is decided by the stage machine, not by judgement

- **Release failed** → the story is held at an in-flight stage, which is exactly what `@session_escalated` covers, and nothing else recovers it: `claimable/2` refuses a later placement (the story is `assigned`, not `contracted`), and `ReclaimExpiredClaimsWorker` is hours away and skips halted tenants.
- **Release succeeded** → the story is back at `queued`, where **there is no escalation edge at all**. So a `revoke` failure cannot be escalated even in principle; its damage is an agent-level credential slot, not this story, and its existing log already names both remedies.
- **Clear** cannot return an error — `{:ok, :cleared | :unchanged}` only, and `:unchanged` is the healthy path.

Two clauses over the release result, **no catch-all**, because a catch-all would silently decline to escalate if a third shape were added.

## Two corrections to my own brief, both read from source

1. **`Stages.human?/1` is not in play.** `human_gate/2` consults it only when `StageMachine.human_only?(edge)`, and that is `edge == :human_resolution` alone (`stage_machine.ex:466`). The empty-`actor_lineage` requirement I warned about does not apply.
2. **The epoch has NOT moved.** `force_unclaim_story/3` does every write in one transaction whose only post-commit step rescues everything and always returns `:ok` (`progress.ex:2977-2991`), so a *reported* release failure is always a rolled-back transaction. The escalation fences on the epoch the story already carries and **deliberately never re-reads it** — a fresh epoch would let this park a story whose claim had gone back and been re-taken by someone else, the hazard `resume/4` already refuses.

## The reason text, and why the remedy leads

Composed entirely by loopctl from constants plus two loopctl-side error terms, so no session-authored text reaches it. **Remedy first, diagnostics last**, because `story_stages_text_bounds` is a CHECK and `Stages.advance/4` refuses an over-long reason *before* the transition — which would leave a story neither placeable nor parked, the exact outcome this exists to prevent.

## Escalation can itself fail, and the recursion stops at one attempt

By argument, not fiat: escalation is the mechanism for reaching a human, there is nothing above a human, and a fallback could only be the same call on the same row with the same failure modes. A refusal logs at `:error` naming the story stranded — the shape `StoryPayload.refuse/4` already uses. A raise is rescued, because the caller is owed the *push* refusal, not this one.

## Evidence

Eight mutations. **One came back exit 1 and the code was deleted rather than shipped**: a whole-text codepoint cap that `mutate.sh` opened to the identity with every test still green, because the existing formatter's limits already bound every term the function can be handed. This module's own rule is that an unreachable clause reading as a guard is worse than no clause, so it went, and the assertion moved to where the bound *is* reachable.

Full suite: **10922 tests, 0 failures.**

## Residual gap, named rather than hidden

**No black-box test drives a genuinely failed release through `place/4`.** Every failure mode of `force_unclaim_story/3` was traced and each breaks the *claim* first, or is unreachable under its own `FOR UPDATE` re-read — which is itself why the incident was a four-hour outlier. The two halves are proved from opposite ends and joined by the mutation that forces the failure on the real undo path. The alternative was a mockable DI seam through the claim path, which is the one path in this module that must not grow a configuration surface.

## Deliberately untouched

846.8 AC-1's step-ordering defect lives in this same function (`clear` runs before `revoke`). It is filed, it is one line, and it wants its own review round rather than riding along in a change about something else. This change adds no new dependence on that ordering.

The review gate is still to run on this branch.


mkreyman:
**The headroom number is wrong a third time, in the same direction, and this time it flips the decision. Measured by me, on this branch, just now.**

| | claimed | measured |
|---|---|---|
| round 1 | "under a third" (~33%) | — |
| round 2 review | ~72% | — |
| round 2 fix | 3643 / 4000 = 91% | — |
| **this measurement** | | **3953 / 4000 = 98.8%, 47 codepoints spare** |

Reproduction — prose with both slots emptied is 727 codepoints, and `inspect(changeset, limit: 5, printable_limit: 200)` on an `%Ecto.Changeset{}` carrying **40 errors AND 40 changes** of 500 chars each is **1613**:

```
prose=727  one_changeset=1613  total=3953  bound=4000  over_by=-47
```

Each previous figure measured a simpler term than the one that actually arrives: a bare 5-key map, then a changeset with `errors` only. `changes` is populated on exactly the paths that produce these errors.

**Why this is not just a stale number.** The round-2 decision to keep the clamp deleted rested on: *"at 91% the hazard that is reachable is this prose growing, and a clamp would absorb exactly that edit silently, where without one the test goes red."* At 98.8% that argument inverts — the prose cannot grow by one word, and the remedy the message exists to deliver is the first thing truncation would eat. Round 2 itself widened this prose by ~63 codepoints to fix the force-unclaim remedy; another correction of that size overruns.

**The fix is the one this PR's own overturn condition already names:** the clamp belongs *inside* `short/1`, not on the composed text where it originally sat — there it also covers `unreleased_claim_event_data/2`'s 8,000-byte bound, which today relies on the same two unbounded-in-one-dimension options.

Folding this into the round-3 fix pass rather than dispatching a second agent at the same file.

mkreyman:
**Correcting my own comment above: my 3953 figure was wrong, and the number I was challenging was right.**

Measured against the term the test **actually builds** — the `%Ecto.Changeset{}` literal at `placement_test.exs:990-1001`, not my approximation of it:

```
prose=727  one_term=1458  total=3643  bound=4000  headroom=357
```

Six fresh VMs reproduce `1458` six times. My `1613` could not be reconstructed from any literal changeset shape; the residual 155 points at one built through `cast/3` with a populated `data` and `types`, which is not what this path receives. So the 91% in the comment was the right figure for the term under test, and my "47 codepoints" overstated the risk.

**What survived the check is a different defect in the same block**, which nobody had re-measured since it was written: "1_080 for a 40-element keyword list" measures **1105**, and "8_631" measures **8831**.

So that block held one number two careful readers could not agree on and one number that was simply stale — which is a better argument for deleting it than either of our figures was. It is deleted in `0f87a8eb`, along with the percentage and the round-by-round archaeology. What stays is checkable: the gate is `reason_within_bound?/1` at 4,000 codepoints of raw text, and the fit is asserted by a named test rather than a line number. The clamp stays deleted because the **assertion** holds the bound, not because a percentage says it fits.

Leaving my wrong comment above rather than editing it, since the reasoning that produced it is the useful part of the record.

PR #866, opened 2026-09-16
A runner's declared capacity is now the capacity loopctl reserves against (846.4)

mkreyman:
## The defect, measured

`runner_pool` showed minis holding `max_sessions: 2` against a declared and configured `1`. **Not stale-since-enrollment**: minis rejoined twice on contract 1.12.0 and the held row did not move either time. The row never followed the declaration.

**It is not cosmetic.** loopctl reserves against the HELD number, so holding 2 against a real 1 places a second concurrent dispatch on a runner that will not run it — and a refused dispatch costs a claim and parks the story, which is exactly the failure #865 just built an escalation for.

## Verified before building, because the fix rests on it

- `max_sessions` is in `RunnerJoin`'s **`required`** list (`runner_contract.ex:573`), so `cast_join/1` refuses a join without it.
- `RunnerStatus` declares only `in_flight`, `draining`, `sample` (`:702-704`) — a status event **can never** carry capacity.

So the declaration arrives exactly once per connection and on every connection. Join is the right place; a backfill would have been wrong.

The update runs **before `Presence.track/4`** (`runner_channel.ex:196`), because `Runners.dispatch/3` needs a live socket in the pool — making that the last instant capacity can change with nobody reserving against the old number.

## One pattern, third instance

The seam is `apply_declaration/3`, not `apply_capacity/3`: `kinds` before contract 1.6.0, `branch_prefixes` in 846.2, capacity now. A runner **declares** its capabilities and loopctl **derives** from the declaration rather than storing an independent copy. `branch_prefixes` is deliberately **not** built here — it is queued behind the runner repo shipping the declaration, and the moduledoc says so.

## The four design questions, answered rather than assumed

**Out of range** cannot reach the row: the contract bounds the wire at 0..64 and `cast_join/1` refuses the rest. The clamp holds for a meta built without that cast, and is bound to the schema's range by a test so a widened column cannot leave it silently refusing values the column accepts.

**Zero** is the interesting case — legal on the wire, illegal in the column. Clamped **up to 1**, not ignored: ignoring keeps the enrolled 2, which is the exact over-reservation being removed. Honouring it needs a migration and makes "send me nothing" a second spelling of `draining`, which already exists and is already enforced by `accepts?/5`. The join is never refused over it, on the same reasoning as `declared_kinds/1`: a machine that cannot get a socket is out of the fleet.

**Lowering below `in_flight`** would have raised on `runners_in_flight_range`. The same statement clamps `in_flight` to `LEAST(in_flight, declared)` — not an invention, it is the answer `Capacity.write_count/4` already gives that state. At `in_flight == max_sessions` the `reserve/3` guard is false, so the machine takes no new work until its live dispatches drain, which is what a runner declaring fewer sessions is asking for. Unreleased ledger rows are untouched and `heal/3` recomputes to the same value.

**The enrollment default** is not dead but is no longer authoritative — documented in the enroll doc, the OpenAPI description and its property, the MCP tool description, the README row and the CHANGELOG. Deliberately **not** made a ceiling: there is no update endpoint for it, so as an operator control it is already broken, and the machine owns the fact. An operator cap is its own change with its own endpoint.

**No audit entry**, argued rather than assumed: enrollment is on the chain because it mints a credential and binds it to a machine; a slot count decides nothing about who implemented, reported or verified anything, it arrives on every reconnect, and the tenant's chain head is the row every writer contends on. The counter-argument and the overturn condition are in the moduledoc verbatim.

## Contract 1.13.0

The `max_sessions` description said loopctl reserves against the value the runner was **enrolled** with "never this value" — this change makes that false, so the version moved and the history row says RE-VENDOR. **Nothing changed on the wire**; no runner has to send anything new. Two MCP tool descriptions became misleading and ship with it.

## Evidence

Six mutations, including removing the call from `:after_join` entirely (5 wiring tests red) and `LEAST` → `GREATEST` on the clamp.

**One came back exit 1 and the TEST was fixed rather than the report.** The cross-tenant test passed without the explicit predicate, because `Repo.with_tenant` had already set the RLS context — so it proved RLS, not the predicate, and the defence-in-depth claim was unfalsifiable. A second call through `AdminRepo`, the only layer at which that predicate is observable, makes it exit 0. **The existing `reserve/3` isolation test has the same shape and is presumably equally inert** — reported, not touched.

10941 tests, 0 failures. mcp-server 678.

## Found and not fixed

- `RunnerJoin.max_sessions` has `minimum: 0` while the column is 1..64. Handled by clamping rather than closing, because tightening the wire would refuse the socket of any runner declaring 0 and drop a machine from the fleet over a number. Closing it means a migration and deciding whether 0 is a second spelling of `draining`.
- Two existing tests silently assumed the enrollment capacity survives a join; their join payloads now declare what they enrol. Any future test that enrols at N and joins with the default silently gets 2.
- There is no loopctl-side remediation for a wrong capacity, by design — the remedy is machine-side. An operator force would partly reverse principle 1.

The review gate is still to run on this branch.


mkreyman:
**Two corrections to this PR body, both of which the review found and one of which I wrote.**

**1. "Unreleased ledger rows are untouched and `heal/3` recomputes to the same value, so nothing here frees a slot a session still holds" — FALSE, and it was the load-bearing safety claim.**

`in_flight` is meant to equal `count(unreleased)`, and releases **decrement** rather than recount. So a runner holding 2 live dispatches that rejoins declaring 1 gets `in_flight: 1` against 2 unreleased rows; the next release takes it to 0 while a session is still running, and `reserve/3` hands out a slot that machine is already using. The drift is **stable**, not transient — `heal/3` computes `min(live, max_sessions)`, which already equals the drifted counter — and `available_runner/2` orders on the understated number, so it *preferentially* selects that machine. The PR reinstated the over-dispatch it was written to end. The reviewer reproduced it.

Fixed in `c2c7dbbb`: releases now recount from `unreleased/2`.

**2. "mcp-server 678" — it is 676**, verified at `HEAD` as well as on the branch. A claim about evidence, so it is worth correcting rather than leaving.

**And one design decision reversed, by my own argument.**

This PR said the enrollment value "seeds the row and does not cap it". That was wrong on the reasoning I gave for it. I argued the drift is asymmetric — over-reserving parks stories, under-reserving only wastes a machine — and then let the declaration win *unconditionally*, which lets a compromised or buggy runner declare 64 and absorb the tenant's admission budget, each dispatch carrying story content and a freshly minted ephemeral key. The asymmetry implies the **minimum**: a machine may always lower itself, never raise itself.

`enrolled_max_sessions` is now its own column, because using the held column as the ceiling would be a one-way ratchet — a machine that once declared 1 would be pinned at 1 for ever.

**Also real, and not something I had considered:** the contract told runner authors that a machine wanting no work sets `draining`, "which loopctl refuses dispatches against". `place_dispatch` did not honour `draining` **and claims the story before pushing**, so following the published advice still lost a claim. That is now enforced on the placement path rather than the sentence being softened.

PR #870, opened 2026-09-16
A runner declares its branch prefixes and loopctl derives a conforming branch (846.2)

mkreyman:
The delivery loop's first real placement was refused `branch_not_allowed`. loopctl derived `feature/story-846-d9975b31` from a hard-coded prefix while the minis runner accepts `loop/` alone, so the dispatch was refused, the story was parked, and the run never started.

**The one field was never the defect.** loopctl chose a prefix, each operator chose a prefix per machine, and nothing reconciled them — an operator could learn the required prefix only by reading a config file on the target box. That is the shape `kinds` had before contract 1.6.0, and it gets the same fix: the runner DECLARES, loopctl DERIVES.

## Contract 1.14.0 — and what re-vendoring does and does not buy

`RunnerJoin.branch_prefixes`, optional, not in `required`. **A runner MUST re-vendor to use this; it need not re-vendor to keep working.** Omitting the field is exactly today's behaviour — no constraint, and the branch is the one loopctl already derived — so an un-upgraded runner is unaffected and nothing on the wire changes for it. That inertness is a TESTED property at three layers, not a claim: the wire (a 1.0.0 join payload is still accepted), the derivation (`branch_for(story, []) == branch_for(story)`, as an equality rather than a literal), and the pushed frame (asserted against `branch_for(story)`, so the claim cannot drift from the derivation).

This is the distinction against the 1.13.0 row, which said re-vendoring was worth it for the description and not required for the wire. There `max_sessions` was already a required field every runner sent and only loopctl's INTERPRETATION changed. Here the field is new, so a runner cannot send it against a vendored 1.13.0 schema — its own pre-flight validation has no such property.

## AC-2 is discharged by DELETION, which is the part worth reading

`DispatchPayload.branch_for/2` is the one derivation. The unattended driver **no longer names a branch at all** — the key was removed from its payload map and `fill/3` fills it. The driver used to call the same function, which was the same name, until the derivation gained an input the driver does not have: the prefix now depends on the live socket the push will reach, and only `Placement` resolves that. A second derivation there would have had to re-read the pool and could still land on a different meta. One call site; nothing to drift.

## The invariant is in the PUBLISHED contract, not our moduledoc

> A RUNNER THAT ENFORCES A PREFIX MUST DECLARE IT: silence is read as no constraint, so a machine enforcing an undeclared prefix refuses every dispatch loopctl sends and each refusal costs the story's claim. Declare the SAME set you enforce — a superset is the same failure.

The test asserts against `RunnerContract.json_schema()` rather than the module's moduledoc, and that is load-bearing: the moduledoc does not travel. What a runner author reads is `priv/runner_contract/v<major>.json`, the file the runner repo vendors, so an invariant stated only in our source binds nobody and cannot be checked from the other side.

## What a prefix may be, and why the wire bound is not enough

This value reaches a git branch name that loopctl hands to a shell on the declaring machine, so a careless or hostile declaration is an injection vector into the control plane's own output.

**The first character must be alphanumeric, and that is the one that matters.** A branch beginning `-` is read by git as an OPTION rather than a ref: `-o` or `--upload-pack=...` turns the runner's own `git checkout` or fetch into argument injection. Anchoring it makes that inexpressible. The class is `A-Za-z0-9`, `_`, `/`, `-` and nothing else, which leaves no shell metacharacter, no whitespace and no control character — `loop/;rm -rf /`, `loop/$(id)` and the backtick form cannot be written. `.` is excluded outright as a deliberate over-restriction: expressing "no `..`" needs a negative lookahead, and a keyword enforced here but absent from the exported set would refuse a join for a reason the runner author cannot read anywhere. `maxLength: 40` and `maxItems: 8` are ONE bound, not two — the value goes verbatim into the Presence meta, which is replicated to every node for the life of the socket and echoed on the pool read, so the resource is entries TIMES length.

**Three gaps the wire cannot close**, so `branch_for/2` re-validates the whole COMPOSED name with `\A...\z`: under PCRE `$` also matches before a final newline, so `"loop/\n"` satisfies the pattern; a per-entry pattern cannot see `loop//` composing into an empty path component; and a Presence meta can be built without passing `cast_join/1` at all.

## When it refuses, and where

`409 no_conforming_branch` with the declared prefixes ECHOED in the body — the echo is the point of the story, since those prefixes were previously readable only by opening a config file on the target machine. 409 rather than 422 because the request is not wrong; it is a state the machine is in.

It refuses the PLACEMENT, never the join. That is 846.4's precedent: a machine that cannot get a socket is out of the fleet entirely, which is far worse than one story that cannot be placed on it. And it refuses rather than CLAMPS, which is where this departs from 846.4 — a smaller number is still a meaningful answer, but there is no conservative branch name. The only way to fit an over-long prefix is to shorten the suffix, and the suffix is the whole of AC-5: it is what keeps two stories on one repository off one branch. A loud refusal beats a quiet collision.

It is checked BEFORE the mint and BEFORE the claim, so nothing is spent: no dispatch row, no ephemeral key, no chain entry, no claim to compensate. The test asserts all of that, not just the error tuple.

A caller-supplied branch outside the declaration is refused `422 branch_not_allowed`, also before the claim. loopctl never REWRITES a caller's branch — silently renaming it would start a session on a name nobody chose, and the operator would go looking for work on the other one.

## A mutation that came back exit 1, and what it found

The AC-2 mutation on the RESUME path PASSED under mutation, so it could not fail: the claim path was covered and the resume path's pass-through was not. A test was written for it — a rejoin declaring different prefixes, then a resume under the same `dispatch_id` — and the re-run is exit 0. That is the wiring-versus-mechanism distinction doing its job, and it was found by mutating rather than by reading.

14 mutations in total. One other returned exit 1 on an unreachable default and is reported as such rather than presented as coverage.

## The first commit is a prerequisite, not part of this story

`0f83b3f4` fixes `derive/1` answering about the repository git was pointed at rather than the path it was asked about. A git hook exports `GIT_DIR`, and `git -C` does not beat it, so the gate running from the pre-commit hook fed the suite an environment naming the invoking repository. The direction hid it: a main checkout exports a RELATIVE `.git` that fails to resolve, so it failed closed by accident, while a linked worktree exports an absolute path that resolves from anywhere and failed open. **Every commit from the main checkout passed and every commit from a worktree was refused**, so this PR could not be committed at all until it landed. Measured harm beyond the red test: `derive/1` on the MAIN tree returned a per-worktree partition, a database name the worktree sweeper treats as disposable. Happy to split it out if you would rather land it on its own.

## Verification

Full suite 10992 tests, 0 failures, verified by this session independently of the agent that wrote the change. `mix credo --strict` clean, dialyzer passed with 0 unnecessary skips, `mix loopctl.check_env_docs` clean, mcp-server 678 tests. `priv/runner_contract/v1.json` regenerated and its two pins updated. The MCP tool ships with the change, per this repo's rule.

Implemented by a dispatched engineer agent; reviewed and independently re-derived by the session before commit. The review gate has not run yet.


mkreyman:
**Round 2 — eight findings, all fixed in `14cecffa`.** Three of them were one design error and the most useful result is a mechanism finding that is not about this PR at all.

## The design error: enumerating fields instead of proving a property

Round 1 closed the argument-injection on `branch` by checking `branch`. Round 2 found `base_branch` open on the adjacent line with an identical `1..255, no pattern` schema, a non-string `branch` skipping the check entirely, and a caller-supplied `branch` defeating the uniqueness the contract publishes. Three leaks, one per round, each closed by naming the next spelling.

`RunnerDispatch` now declares `@ref_fields [branch: :story_unique, base_branch: :shared]` and `@non_ref_string_fields [:dispatch_id, :story_id, :kind, :repo]` — a TOTAL partition of the schema's string properties — and `DispatchPayload.validate_refs/2` iterates that declaration. It runs twice: on the caller's map, so a bad value costs nothing, and on the FINISHED payload, so a `base_branch` filled from the intake source is covered too. Both strictly before the claim. A contract test fails in both directions when a new string property appears.

**Its bound, stated rather than left to be discovered:** the test forces an author to classify a new field; it cannot catch one deliberately misclassified as non-ref. No mechanical test for "does this string reach a git ref argument" exists, and a name heuristic (`~r/branch|ref/`) was rejected — it would miss a field called `target` and would read as stronger than it is. The four current non-ref fields are each independently closed by a `format`, `enum` or `pattern`.

## DIALYZER ONLY REFUSES AN IMPOSSIBLE CONTRACT, NEVER A TOO-NARROW ONE

This is the finding worth carrying out of the PR. Finding 5 was `Placement`'s `@type error()` omitting a member the code returns. The first fix relied on dialyzer to catch a recurrence — and mutating the union away left `mix dialyzer` GREEN. A typespec that is narrower than the code's actual returns is not a contradiction, so nothing refuses it. **The typespec had no checker at all**, which is exactly how round 1 dropped a member in the first place and why nobody noticed.

It now ships with a test that reads both `@type error()` unions off the compiled typespecs and asserts the subset relation. Re-mutated against that test: exit 0.

That generalises well past this file: anywhere a `@type` or `@spec` is treated as a guarantee, dialyzer is not the thing guaranteeing it.

## The rest

**Resume no longer strands a live claim.** A prefix is a fact about ANOTHER MACHINE whose remedy is a config file on that box; a name is a fact about the string in THIS REQUEST, retryable at once because the refusal writes nothing. Resume is exempt from the first and not the second.

**A resume re-pushes the name it first derived.** `runner_dispatches.branch` (nullable, no backfill) records it at the first push. This restores a property that existed before this PR — the branch used to be a pure function of the story and could not move across retries. Since a ledger row at `sent` means only that no reply was RECORDED, a session may already be running on the first branch, so merely documenting the hazard would have meant documenting that we may push a second branch at a machine already working the first. A retry naming a different branch is refused `branch_conflict` rather than silently substituted, and `validate_refs/2` runs before the pin so `-o` reads as malformed rather than as a conflict.

**A misconfigured runner no longer halts a repo.** The driver advances to the next candidate on `no_conforming_branch` and on nothing else — safe because that refusal is decided before anything is claimed or minted — and still reports `:blocked` when every machine is misconfigured.

**Finding 8 was decided by probe rather than by agreement between the two of us.** git 2.53.0 ignores `core.worktree` both from a `GIT_CONFIG_GLOBAL` file and from `GIT_CONFIG_COUNT/KEY/VALUE` injection, so config cannot move what the derivation answers — while clearing `GIT_CONFIG_GLOBAL` on a box configured through it loses `safe.directory`, which silently returns the worktree suite to the shared database. So `GIT_CONFIG*` is excepted alongside `GIT_TRACE*`, and the doc's old closing claim that a non-zero exit is "the safe one" is corrected, because here it is backwards. Unlike the `GIT_TRACE*` exception this one is falsifiable and ships with three tests, one asserting the premise against raw git.

## Behaviour change worth knowing before it is found in production

An existing intake source carrying an invalid `base_branch` is now refused at placement instead of being pushed to a machine. Intended — that value reaches a git command on the runner — but it is a change for an existing bad row.

## Verification

Thirteen mutations, twelve exit 0. **The thirteenth came back exit 1 and changed the fix** rather than being explained away — that is finding 5 above. Full suite 11024 tests, 0 failures; `mcp-server` 678. `priv/runner_contract/v1.json` byte-unchanged, so no version bump and no re-vendor.

